PR #3, opened 2024-04-29
README.md: Drop "BACKEND" from the title

russellb:
This code is not backend-specific, so drop it from the readme title.
While we're at it, capitalize DeepSpeed as the upstream project does
it.


russellb:
@Maxusmusti did you close this on purpose?

russellb:
Nevermind - I see this happened when the repo was made public. I can repost it.

PR #4, opened 2024-04-30
add conversion script to format sdg output

jaideepr97:
adds script to convert generated data to match the output of backed SDG


Maxusmusti (approved):
Amazing, beautifully written

PR #5, opened 2024-04-30
README.md: Drop "BACKEND" from the title

russellb:
This code is not backend-specific, so drop it from the readme title.
While we're at it, capitalize DeepSpeed as the upstream project does
it.


russellb:
re-post of #3

russellb:
included in #7

PR #6, opened 2024-05-01
feat: automatically pick latest train data file

jaideepr97:
This change allows the script to accept a train data dir as input and automatically picks the latest train data file based on the ISO8601 timestamp in the name of the file


ilab_to_sdg.py
```diff
@@ -1,17 +1,32 @@
 import json
 import sys
 from pathlib import Path
+import glob
+import os
+import re
+from datetime import datetime
```
russellb, line 7: I don’t think you’re using datetime or re anymore

russellb (commented):
One last tweak then lgtm!

ilab_to_sdg.py
jaideepr97: fixed, thanks!

russellb (approved):

Maxusmusti (approved):
LGTM

PR #10, opened 2024-05-09
Add CPU offloading capabilities

kdvalin:
This PR allows for Deepspeed to offload some GPU compute requirements to the CPU.  This drastically lowers required VRAM to run the training backend by moving data to system memory.  With this, training can be run on a single GH200 with 96 GiB of VRAM.

More information: https://www.deepspeed.ai/tutorials/zero-offload/


n1hility:
Thank you for this! LGTM 

PTAL @Maxusmusti

mnmehta:
Thanks Keith. For now I think it is sufficient to just provide the offload_optimizer and offload_param options, the deepspeed_optimizer should be picked automatically ... no optimizer offload => FusedAdam and cpu optimizer offload => DeepSpeedCPUAdam. Otherwise we'll need more error checking, e.g. using cpu optimizer offload with FusedAdam optimizer is not supported.

main_ds.py
```diff
@@ -19,23 +19,23 @@
...
         "train_batch_size": samples_per_gpu * world_size * grad_accum,
         "gradient_accumulation_steps": grad_accum,
         "train_micro_batch_size_per_gpu": samples_per_gpu,
         "steps_per_print": 1,
         "zero_optimization": {
             "stage": 2,
-            "offload_param": {"device": "none"},
-            "offload_optimizer": {"device": "none"},
+            "offload_param": {"device": offload_param},
+            "offload_optimizer": {"device": offload_optimizer},
```
markmc, line 38: We're using ZeRO-2 but parameter offloading is [only available in ZeRO-3](https://www.deepspeed.ai/docs/config-json/#parameter-offloading)? Sounds like `offload_param` won't actually do anything with `stage = 2`?

aldopareja:
we have another PR coming https://github.com/instructlab/training/pull/12 that will add offloading more robustly. Feedback there would be appreciated!

main_ds.py
fabianlim: that is correct, I woldnt allow `"offload_param"` to be configurable here for that reason.

main_ds.py
```diff
@@ -270,6 +276,8 @@ def main(args):
     )
     parser.add_argument("--is_granite", action="store_true")
     parser.add_argument("--max_batch_len", type=int, default=60000)
+    parser.add_argument("--offload_optimizer", type=str, default="none", choices=["none", "cpu"])
+    parser.add_argument("--offload_param", type=str, default="none", choices=["none", "cpu"])
```
fabianlim, line 280: we shouldnt allow this. It is not supported by zero2

PR #12, opened 2024-06-11
Create interface for accessing the training code

RobotSail:
This PR turns this into a library.

View the following PR for an example: https://github.com/instructlab/instructlab/pull/1329

The main idea is that we provide a function like:
```py
def train_torchrun(torchargs: TorchrunArguments, training_args: FullTrainingArguments):
  pass
``` 

Each class `TorchrunArguments` and `FullTrainingArguments` provides training-specific arguments. And then from another library, you would simply provide these arguments to the `train_torchrun` function. 

For any other training method that we define, we could provide a similar interface depending on which arguments are needed.

We separate the arguments here because `TorchrunArguments` are the ones passed to `torchrun` and then the full training arguments are the ones that we actually train with. It's not crucial that these are different, but it makes our lives a lot easier from a maintenance standpoint.


setup.py
```diff
@@ -0,0 +1,22 @@
+import os
```
russellb, line 1: Instead of setup.py, can you create a pyproject.toml similar to the `instructlab` and `sdg` repos?

The `sdg` one is probably a better one to copy, since it's a library similar to this one.

ilab_train/config.py
```diff
@@ -0,0 +1,47 @@
+from dataclasses import dataclass
```
russellb, line 1: To be consistent with other libraries, can you set this up as `instructlab.train` as the library?

See this PR for some guidance on what to change to do this:

https://github.com/instructlab/sdg/pull/14

russellb (changes requested):

setup.py
RobotSail: Sure, I'm not married to `setup.py` and find it to be clunky. Seems like pyproject.toml is a better solution.

ilab_train/config.py
RobotSail: Sure, I like that more.

setup.py
nathan-weinberg: I copied the `sdg` config for https://github.com/instructlab/eval - all three libs should work the same and remain in sync

ilab_train/config.py
nathan-weinberg: @RobotSail you can also reference https://github.com/instructlab/eval/pull/1

RobotSail:
https://github.com/instructlab/eval/pull/1

pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+    "License :: OSI Approved :: Apache Software License",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: POSIX :: Linux",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
russellb, line 28:
```suggestion
    "Programming Language :: Python :: 3",
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.11",
    "Programming Language :: Python :: 3.12",
```

pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+[tool.setuptools.packages.find]
+where = ["src"]
+include = ["instructlab.training"]
+
+[tool.setuptools.package-data]
+"instructlab.schema.v1" = ["*.json"]
+"instructlab.schema.v2" = ["*.json"]
+
+[tool.ruff]
+target-version = "py39"
```
russellb, line 60:
```suggestion
target-version = "py310"
```
```

File: pyproject.toml [hunk 5]
```diff
@@ -0,0 +1,108 @@
...
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[tool.setuptools.packages.find]
+where = ["src"]
+include = ["instructlab.training"]
+
+[tool.setuptools.package-data]
+"instructlab.schema.v1" = ["*.json"]
+"instructlab.schema.v2" = ["*.json"]
```
russellb, line 57:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+]
+dynamic = ["dependencies", "optional-dependencies", "version"]
+
+[project.scripts]
+
+[project.urls]
+homepage = "https://instructlab.io"
```
russellb, line 36:
```suggestion
homepage = "https://instructlab.ai"
```

requirements.txt
```diff
@@ -0,0 +1,10 @@
+pyyaml
+py-cpuinfo
+torch>=2.2.0
+transformers
+datasets
+numba
+numpy
+rich
+
+deepspeed
```
russellb, line 10: Not for this PR, but later this should get updated to add some more specific versions for each line.

Take a look at requirements.txt in the `instructlab/instructlab` repo for some examples where we put guardrails around minimum/maximum for each

russellb (commented):
I haven't reviewed the code part, just pyproject/requirements/namespace package portions. That all looks good, though I gave some suggested edits for some minor cleanup. That includes dropping Python 3.9 since I think we landed on 3.10 as the minimum.

pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+
+[project]
+name = "instructlab-training"
+authors = [
+    { name="InstructLab", email="dev@instructlab.ai" },
+]
+description = "Training Library"
+readme = "README.md"
+license = {text = "Apache-2.0"}
+requires-python = ">=3.9"
```
nathan-weinberg, line 15:
```suggestion
requires-python = ">=3.10"
```

requirements.txt
nathan-weinberg: +1 to this, also the empty line L9 should be removed

@mairin has highlighted that torch 2.2.0 has a CVE, so can we bump that to 2.3.0?

nathan-weinberg (changes requested):
Just focused on the Python packaging stuff, will leave the training code itself to y'all for now :smile:

requirements.txt
russellb: This actually raises an important point of pain we're going to hit immediately by splitting up into these separate repos.

All of these requirements.txt files have to be compatible. We'll need a process for dealing with any and all requirements changes to ensure it's done consistently across repos to ensure they remain compatible.

RobotSail: I agree, we should add the specific requirements down the line. They probably shouldn't be added in this PR yet while we're still sorting things out.

As for the empty line, we need deepspeed to be the very last entry since it depends on the ones above and breaks if it's included anywhere other than here.
I'll add a comment to explain this.

pyproject.toml
RobotSail: Thanks for this suggestion!

pyproject.toml
RobotSail: Thanks!

pyproject.toml
RobotSail: Good catch, thanks!

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +413,123 @@ def main(args):
...
+        command = [
+            "torchrun",
+            f"--nnodes={torch_args.nnodes}",
+            f"--node_rank={torch_args.node_rank}",
+            f"--nproc_per_node={torch_args.nproc_per_node}",
+            f"--rdzv_id={torch_args.rdzv_id}",
+            f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
+            __file__,
+            f"--model_name_or_path={train_args.model_name_or_path}",
+            f"--data_path={train_args.data_path}",
```
RobotSail, line 447: Mustafa: this should be the processed data path

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +413,123 @@ def main(args):
...
+            f"--output_dir={train_args.output_dir}",
+            f"--num_epochs={train_args.num_epochs}",
+            f"--effective_batch_size={train_args.effective_batch_size}",
+            f"--learning_rate={train_args.learning_rate}",
+            f"--num_warmup_steps={train_args.num_warmup_steps}",
+            f"--save_samples={train_args.save_samples}",
+            f"--log_level={train_args.log_level}",
+            f"--max_batch_len={train_args.max_batch_len}",
+            f"--seed={train_args.seed}",
```
RobotSail, line 456: Mustafa: needs to add the new args for lora & quantization

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +413,123 @@ def main(args):
...
+        print(f"\033[92mRunning command: {' '.join(command)}\033[0m")
+
+        with open("logfile.out", "w", encoding="utf-8") as logfile:
+            subprocess.run(
+                command,
+                stdout=logfile,
+                stderr=subprocess.STDOUT,
+            )
+
+        # stream the stdout and stderr output
```
RobotSail, line 476: need to implement this

Maxusmusti:
Note: Current requirements will overwrite existing nvidia pytorch installs. Need to ensure that if those exist, we are not installing our own torch.

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
     torch.distributed.destroy_process_group()
 
 
+def run_training(torch_args: TorchrunTrainArgs, train_args: FullTrainArgs):
```
Maxusmusti, line 415: Update names of imports

src/instructlab/training/config.py
```diff
@@ -0,0 +1,107 @@
...
+    # Either the name of a HuggingFace model or a path to a model saved in HuggingFace format.
+    model_path: str
+
+    # this field specifies the filepath to the training dataset before processing
+    data_path: str
+    ckpt_output_path: str
+
+    # this field defines where we should be saving the processed version of the training dataset
+    # after we have tokenized it
+    processed_data_output_path: str
```
Maxusmusti, line 91: Change to directory, name doesn't matter

src/instructlab/training/config.py
```diff
@@ -0,0 +1,107 @@
...
+
+    num_gpus: int
```
Maxusmusti, line 93: to be removed (already in torchrun args)

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+            # An important reason for why #1 would be preferable is in the case of OpenShift/SELinux
+            # where the user has a defined place for new temporary data to be written.
+            data_output_path=train_args.processed_data_output_path,
+            model_path=train_args.model_path,
+            data_path=train_args.data_path,
+            max_seq_len=train_args.max_seq_len,
+        )
+    )
+
+    if not os.path.exists(train_args.output_dir):
```
Maxusmusti, line 435: Change to ckpt_output_dir

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+            "torchrun",
+            f"--nnodes={torch_args.nnodes}",
+            f"--node_rank={torch_args.node_rank}",
+            f"--nproc_per_node={torch_args.nproc_per_node}",
+            f"--rdzv_id={torch_args.rdzv_id}",
+            f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
+            __file__,
+            f"--model_name_or_path={train_args.model_name_or_path}",
+            f"--data_path={train_args.processed_data_output_path}",
+            f"--output_dir={train_args.output_dir}",
```
Maxusmusti, line 448: fix here too

src/instructlab/training/main_ds.py
Maxusmusti: *path

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+    try:
+        command = [
+            "torchrun",
+            f"--nnodes={torch_args.nnodes}",
+            f"--node_rank={torch_args.node_rank}",
+            f"--nproc_per_node={torch_args.nproc_per_node}",
+            f"--rdzv_id={torch_args.rdzv_id}",
+            f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
+            __file__,
+            f"--model_name_or_path={train_args.model_name_or_path}",
```
Maxusmusti, line 446: change to just model_path

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+            f"--data_path={train_args.processed_data_output_path}",
+            f"--output_dir={train_args.output_dir}",
+            f"--num_epochs={train_args.num_epochs}",
+            f"--effective_batch_size={train_args.effective_batch_size}",
+            f"--learning_rate={train_args.learning_rate}",
+            f"--num_warmup_steps={train_args.num_warmup_steps}",
```
Maxusmusti, line 452: change to just warmup_steps

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+            f"--save_samples={train_args.save_samples}",
+            f"--log_level={train_args.log_level}",
```
Maxusmusti, line 454: change to just "INFO"

src/instructlab/training/config.py
```diff
@@ -0,0 +1,107 @@
...
+    max_seq_len: int
+    max_batch_len: int
+    num_epochs: int
+    effective_batch_size: int
+    save_samples: int
+    learning_rate: float
+    warmup_steps: int
+
```
Maxusmusti, line 101: add seed as option

src/instructlab/training/config.py
```diff
@@ -0,0 +1,107 @@
...
+    ds_offload_strat: DeepSpeedOffloadStrategy
```
Maxusmusti, line 102: add mock_data as option

src/instructlab/training/config.py
```diff
@@ -0,0 +1,107 @@
...
+    processed_data_output_path: str
+
+    num_gpus: int
+    max_seq_len: int
+    max_batch_len: int
+    num_epochs: int
+    effective_batch_size: int
+    save_samples: int
+    learning_rate: float
+    warmup_steps: int
```
Maxusmusti, line 100: add is_granite (but maybe as is_padding_free)

src/instructlab/training/main_ds.py
```diff
@@ -387,16 +412,101 @@ def main(args):
...
+        command = [
+            "torchrun",
+            f"--nnodes={torch_args.nnodes}",
+            f"--node_rank={torch_args.node_rank}",
+            f"--nproc_per_node={torch_args.nproc_per_node}",
+            f"--rdzv_id={torch_args.rdzv_id}",
+            f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
+            __file__,
+            f"--model_name_or_path={train_args.model_name_or_path}",
+            f"--data_path={train_args.processed_data_output_path}",
```
Maxusmusti, line 447: expected hera as file, but in process_data as a dir

README.md
```diff
@@ -1,3 +1,238 @@
...
+
+
+```python
+# on machine 2
+torchrun_args = TorchrunArgs(
+    nnodes = 2, # number of machines 
+    nproc_per_node = 4, # num GPUs per machine
+    node_rank = 1, # node rank for this machine
+    rdzv_id = 123,
+    rdzv_endpoint = '127.0.0.1:12345'
```
JamesKunstle, line 231: very minor nit- the second machine wouldn't refer to localhost, it'd refer to the public IP of the main machine.

src/instructlab/training/config.py
```diff
@@ -0,0 +1,127 @@
...
+    # https://github.com/instructlab/training/issues/26
+    # NVME = "nvme"
+    NONE = None
+
+
+class QuantizeDataType(Enum):
+    """
+    Defines what datatype we use during quantization.
+    """
+
```
JamesKunstle, line 26: should there be an 8-bit quanitzation size that's supported?

JamesKunstle (changes requested):
two thoughts, one being a minor nit-pick, but in general this is very good, I'll check that it runs.

src/instructlab/training/config.py
Maxusmusti: Not yet, we can experiment with and include that in a follow-up once we've got some validation for it (if we have time). For now our code only has 4-bit support, so makes sense for interface to match.

RobotSail: Yes but the LoRA training which Fabian implemented only supports 4-bit quantization. Since this PR is only concerned with the interface, we should leave this for a follow-up PR.

Made an issue for it here:
https://github.com/instructlab/training/issues/29

README.md
RobotSail: Good catch!

RobotSail: Updated

RobotSail:
Thank you for the review @JamesKunstle, I've created an issue about your quantization comment here: https://github.com/instructlab/training/issues/29

src/instructlab/training/config.py
```diff
@@ -0,0 +1,125 @@
...
+    Defaults are all taken from the above docs.
+    """
+
+    cpu_offload_optimizer: bool = False
+    cpu_offload_optimizer_ratio: float = 1
+    cpu_offload_optimizer_pin_memory: bool = False
+
+
+class TrainingArgs(BaseModel):
```
cdoern, line 85: PTAL at instructlab/instructlab#1370

src/instructlab/training/main_ds.py
```diff
@@ -7,34 +7,55 @@
...
+    TrainingArgs,
+    TorchrunArgs,
+    DataProcessArgs,
+    DeepSpeedOptions,
 )
+import instructlab.training.data_process as dp
 
 
-def get_ds_config(world_size, samples_per_gpu, grad_accum):
+def get_ds_config(world_size, samples_per_gpu, grad_accum, opts: DeepSpeedOptions):
```
cdoern, line 50: ok, I have been operating under the assumption that the Ds config would live in the cli. What if eventually we want to allow ppl to give their own ds config?

src/instructlab/training/main_ds.py
```diff
@@ -223,9 +283,10 @@ def maybe_resume_training(args, model):
     except FileNotFoundError:
         pass
 
-    # we will update the start step here 
+    # we will update the start step here
     return model
-    
+
```
cdoern, line 288: two blank lines

src/instructlab/training/main_ds.py
```diff
@@ -404,16 +469,111 @@ def main(args):
     torch.distributed.destroy_process_group()
 
 
+def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs):
```
cdoern, line 472: so, is the cli meant to call this? In terms of passing the args would it be possible to have torch and train args as one object? otherwise if we ever let the user customize this, would they have to understand the difference btwn the two?

cdoern: also,  I may just be missing this, but where is the preserved CLI linux train code?

cdoern (commented):
initial comments. Want to make sure this is aligned with CLI impl.

src/instructlab/training/main_ds.py
RobotSail: We still need you to pass the DeepSpeedConfig, you would just be doing it via the `deepspeed` field in `TrainingArgs`. the `get_ds_config` option just builds the actual config that deepspeed itself consumes. 

The reason we're doing it this way is to limit the surface of what we enable with DeepSpeed.

src/instructlab/training/main_ds.py
RobotSail: Yes the CLI will be calling this. 

> also, I may just be missing this, but where is the preserved CLI linux train code?

We haven't moved over the existing CLI train code and instead added a LoRA implementation that @fabianlim provided us.

src/instructlab/training/main_ds.py
RobotSail: Formatter made this change.

src/instructlab/training/config.py
JamesKunstle: ++ sounds great.

src/instructlab/training/main_ds.py
Maxusmusti: the training library has full feature parity with the old linux train code, lora/qlora supported

cdoern (approved):

russellb (changes requested):
Can you squash the commit history here before merging?

README.md
```diff
@@ -1,3 +1,245 @@
...
+
+# or, for an editable install (development)
+pip install -e .[cuda]
+```
+
+
+## Usage
+
+Using the library is fairly straightforward, simply import the necessary items,
+namely:
```
russellb, line 42: Minor suggestion:

```suggestion
To use the library, import the necessary items:
```

README.md
```diff
@@ -1,3 +1,245 @@
...
+	learning_rate = 2e-6,
+	warmup_steps = 800,
+	is_padding_free = True, # set this to true when using Granite-based models
+	random_seed = 42,
+)
+```
+
+
+We'll also need to define the settings for running a multi-process job
+via `torchrun`. To do this, simply create a `TorchrunArgs` object.
```
russellb, line 80: Minor suggestion:

```suggestion
via `torchrun`. To do this, create a `TorchrunArgs` object.
```

JamesKunstle (approved):
v good thank you for all your work

requirements.txt
```diff
@@ -0,0 +1,26 @@
...
+pydantic>=2.7.0
+# TODO: add dolomite_engine as a dependency
+# for the time being, you can install it using this command:
+# pip install git+https://github.com/ibm-granite/dolomite-engine.git
+# Note: this doesn't install its dependencies, for those you should install the ones
+# in this list: https://github.com/ibm-granite/dolomite-engine/blob/main/requirements.txt
+# 
+
+# deepspeed needs to be at the bottom or it'll break during installation
+deepspeed==0.14.3
```
russellb, line 25: Can you file a follow-up issue about restricting the versions and ensuring they're compatible with `instructlab/instructlab` ?

nathan-weinberg (approved):

src/instructlab/training/config.py
```diff
@@ -0,0 +1,132 @@
...
+    All the arguments consumed by the training data pre-process script.
+    """
+
+    data_path: str
+    data_output_path: str
+    max_seq_len: int  # defines the max sequence length of a sample
+    model_path: str  # either a HF model name or path to HF model
+
+    # disable the protected namespace for the model_config field
+    model_config = ConfigDict(protected_namespaces=())
```
russellb, line 44: These are all class variables, right?

I would think having one instance of every one of these for all instances of the class isn't really what you want? This applies to most of the classes after this one.

russellb (changes requested):
biggestion question here is about the use of class variables.

In terms of PR structure -- it would have been nice to have the PR that's doing the library setup setup from all the API introduction stuff.

Maxusmusti (approved):
LGTM, tested full training (no lora/offload) with both base scripts and updated interface, both functioning as expected

README.md
RobotSail: I love it

README.md
RobotSail: 👨‍🍳 🤌

requirements.txt
RobotSail: In this repo or the CLI one?

src/instructlab/training/config.py
RobotSail: No, see the docs for Pydantic (and dataclasses):
https://docs.pydantic.dev/latest/

It lets you create type-safe structs, but you declare the variables at the class-level. It's unusual, but it works.

requirements.txt
russellb: this one

src/instructlab/training/config.py
russellb: Thanks for clarifying!

setup.py
```diff
@@ -0,0 +1,17 @@
...
```
russellb: looks like you can drop this commit, right?  or squash it

.gitignore
```diff
@@ -0,0 +1,166 @@
+# Attribution:
```
russellb, line 1: You can squash this into the first commit.

When do you do so, can you update the commit message so it's a single message that describes the sum of what's included?

russellb:
just to be super clear:

1. squash the first two commits and fix up the commit message
2. create a follow-up issue for cleaning up requirements.txt to specify versions for all dependencies + making sure the versions are aligned with `instructlab/instructlab` if it's a shared dependency

setup.py
RobotSail: oh right we can just drop this

.gitignore
RobotSail: Yep

russellb:
> 2. create a follow-up issue for cleaning up requirements.txt to specify versions for all dependencies + making sure the versions are aligned with `instructlab/instructlab` if it's a shared dependency

We talked about this, so I know your intent was to create this issue, but I can't find it anywhere. I went ahead and filed this one: https://github.com/instructlab/training/issues/34

PR #15, opened 2024-06-13
Add flag for disabling flash attention

Maxusmusti:
For hardware scenarios where flash attention is potentially not supported.

### DIFF
--- a/main_ds.py
+++ b/main_ds.py
@@ -57,21 +57,23 @@ def setup_model(args, tokenizer, train_loader, grad_accum):
             bnb_4bit_compute_dtype=torch.float16,  # if not set will throw a warning about slow speeds when training
         )
         
+    base_model_args = {
+        "pretrained_model_name_or_path": args.model_name_or_path,
+        "torch_dtype": torch.bfloat16,
+        "quantization_config": bnb_config,
+    }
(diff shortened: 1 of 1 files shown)


aldopareja:
are we sure that not specifying the attention mechanism is enough for this not to fail?. We need to setup the 3090 machine from Mo, I guess we can spinup docker containers there just to not mess up with her drivers?

main_ds.py
```diff
@@ -57,21 +57,23 @@ def setup_model(args, tokenizer, train_loader, grad_accum):
...
+        base_model_args["attn_implementation"] = "flash_attention_2"
+
     if args.is_granite:
         from dolomite_engine.hf_models.models import GPTDolomiteForCausalLM
         model = GPTDolomiteForCausalLM.from_pretrained(
-            args.model_name_or_path,
-            attn_implementation="flash_attention_2",
-            torch_dtype=torch.bfloat16,
+            **base_model_args,
             use_padding_free_transformer=True,
```
fabianlim, line 72: if you want to disable FA2, you also need to disable padding free, since it uses FA2 under the hood

RobotSail (commented):
Will need to test this but otherwise it looks good.

PR #18, opened 2024-06-14
enable model to be save every epoch ends; configurable with a parameter

dushyantbehl:


aldopareja:
it's better if we only save every number of steps, otherwise we could end up saving checkpoints too close to each other. Although we can keep this branch for debugging.

PR #19, opened 2024-06-14
Multipack padding fix

Maxusmusti:
For non-dolomite models, packing_max_batch_len was being calculated incorrectly. This calculates the necessary increase to make sure that average batch size is similar to the specified effective_batch_size.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -462,7 +462,9 @@ def main(args):
         effective_batch_size=args.effective_batch_size,
         max_batch_len_per_gpu=args.max_batch_len,
         is_padding=not args.is_granite,
-        dataset_lengths=dataset.get_lengths()
+        dataset=dataset,
+        pad_id=tokenizer.pad_token_id,
+        seed=args.seed
     )
     args.samples_per_gpu = (
         args.effective_batch_size // grad_accum // torch.distributed.get_world_size()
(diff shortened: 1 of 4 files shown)


aldopareja (approved):

russellb:
@aldopareja @Maxusmusti I have a couple minor concerns with this merge process-wise:

1. It looks like the lint CI job was failing. Ideally nothing should merge with a failing job.
2. The commit history is not in an ideal state for including in the repo. This is OK while you're in development, but merging all of these "fixes" commits isn't ideal. In this case, I would suggest squashing this all down to a single commit before merging.

Maxusmusti:
Hey @russellb those are fair concerns. In this case:

 - the only lint failure was for an erroneous check that was removed from the repo, all of the formatting / linting checks passed

 - the PR was squashed on merge, so it is represented as a single commit with a more proper name / message in our history

russellb:
> Hey @russellb those are fair concerns. In this case:
> 
> * the only lint failure was for an erroneous check that was removed from the repo, all of the formatting / linting checks passed

ok, thanks.

> * the PR was squashed on merge, so it is represented as a single commit with a more proper name / message in our history

Ah, I didn't notice that. That helps in this case.

However, note that the org-wide policy is to use merge commits. That does put more responsibilities on maintainers to ensure there's an intentionally structured commit history in a PR before merging it.  See this doc for more:

https://github.com/instructlab/dev-docs/blob/main/docs/github-merge-strategy.md

so we should be turning off the other merge methods at some point

PR #20, opened 2024-06-14
create trainer class(es) from `main_ds.py`

JamesKunstle:
Reorganizes the training code currently in `main_ds.py` into objects.

## Adds:

* DeepSpeedTrainer
* DeepSpeedModelWrapper
* MultipackDataWrapper

## Subjects for review:

1. Inputs to constructors of objects are _exhaustive_ intentionally- I wanna be smart about how we design it. How do we want to wrap these up into configuration objects? Do we want to reuse the config objects from the `config` objects that Oleg's PR put together? Do we want our own?
2. How do we expect for these objects to be consumed? These are pretty tightly coupled to one another, but one could use a different *DataWrapper object for a different sample strategy, so they're not *totally* coupled.
3. How else could we break these objects up?
4. Exhaustively define the parameters so contributors / users know which to attend to vs. which to ignore.


JamesKunstle:
From `mistral-finetune`: 

```
The goal of this repository is to provide a simple, guided entrypoint to finetune Mistral models. As such, it is fairly opinionated (especially around data formatting) and does not aim at being exhaustive across multiple model architectures or hardware types. For more generic approaches, you can check out some other great projects like [torchtune](https://pytorch.org/torchtune/stable/overview.html).
```

I think we should adopt the same design philosophy. We should aim to accept a subset of model architectures that we've verified (e.g. granite via dolomite, llama, mistral) with opinionated optimizations on data sampling throughput. e.g. because of Multipack, we have to do distributed loss reduce-sum.

Then we can train in the context of the distributed training framework (DeepSpeed vs. FSDP) with a specific trainer class.

src/instructlab/training/trainer.py
```diff
@@ -112,9 +113,9 @@ def __init__(
     ):
         # NOTE: this is exhausive. In review, let's decide how to split this up into config groups.
         self.model_name_or_path: str = model_name_or_path
-        self.data_loader: any = data_loader 
+        self.data_loader: Any = data_loader
```
RobotSail, line 116: Does setting these as `Any` make it so that subsequent references are typed? Or will they appear as `any`.

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+    1. Loads model, accepts quantization / LoRA configuration.
+    2. Initializes optimizer, model, learning-rate scheduler for DeepSpeed distributed training.
+    """
+
+    def __init__(
+        self,
+        model_name_or_path: str,
+        data_loader: Any,  # TODO: type the DataLoader obj.
+        output_dir: str,
+        tokenizer: Any,  # TODO: type the tokenizer obj.
```
RobotSail, line 96: The type is `PreTrainedTokenizer`

JamesKunstle: ++

src/instructlab/training/trainer.py
JamesKunstle: it's because the variables seem naked without a type

fabianlim:
Just one comment maybe file should be called `trainer_ds.py`?

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+def main(_args: argparse.ArgumentParser):
+    if os.environ["LOCAL_RANK"] == "0":
+        print(f"\033[38;5;120m{yaml.dump(vars(_args), sort_keys=False)}\033[0m")
+
+    setup_logger(_args.log_level)
+
+    distributed_init(_args)
+
+    dataw = MultipackDataWrapper(
+        model_name_or_path=_args.model_name_or_path,
```
fabianlim, line 533: it doesnt seem very intuitive that data pre requires the model name. It seems more intuitive to pass the tokenizer. But this means the model has to be loaded outside.

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+        data_path=_args.data_path,
+        effective_batch_size=_args.effective_batch_size,
+        max_batch_len=_args.max_batch_len,
+        world_size= int(os.getenv("WORLD_SIZE")),
+        is_padding_free=_args.is_granite,
```
fabianlim, line 538: this introduces some coupling into the DataWrapper, since it actually doesnt really need to know if the model is granite or not. it just needs some dependency injection

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+
+def main(_args: argparse.ArgumentParser):
+    if os.environ["LOCAL_RANK"] == "0":
+        print(f"\033[38;5;120m{yaml.dump(vars(_args), sort_keys=False)}\033[0m")
+
+    setup_logger(_args.log_level)
+
+    distributed_init(_args)
+
+    dataw = MultipackDataWrapper(
```
fabianlim, line 532: I feel that you dont really need an object for this. All that is really needed is a dataloader being prepared.

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+            f"effective_batch_size: {_args.effective_batch_size}\n"
+            f"max_batch_len_per_gpu: {_args.max_batch_len}\n"
+            f"packing_max_batch_len: {dataw.packing_max_batch_len}\n"
+            f"grad_accum: {dataw.grad_accum}\n"
+            f"num batches: {len(dataw.data_loader)}\n"
+            f"avg_samples_per_batch: {len(dataw.dataset)/len(dataw.data_loader)}\n"
+            f"samples_per_gpu: {dataw.samples_per_gpu}\033[0m"
+        )
+
+    modelw = DeepSpeedModelWrapper(
```
fabianlim, line 555: similarly here, all you need is a model being prepared.

src/instructlab/training/trainer.py
JamesKunstle: This was a circular dependency that I'm glad you noticed. We can load the tokenizer independently, which I think breaks the cycle. The model setup step requires information calculated in the DataWrapper class. We could split model and tokenizer loading out from DeepSpeedModelWrapper, which I think woudl break up the parts better. What do you think?

src/instructlab/training/trainer.py
JamesKunstle: The DataWrapper _does_ need to know about whether the model is padding free because the MultipackDataSampler uses that information for batch allocation w.r.t padding. Do you intuit that there's a better split for this class? The sample could be initialized elsewhere.

src/instructlab/training/trainer.py
JamesKunstle: Very valid feedback. I wrapped these things into a pseudo-dataclass because they all related downstream, and the calculations required each other in order. It's more a Wrapper than a proper class in that way.

src/instructlab/training/trainer.py
JamesKunstle: As before, these data prep and model prep steps _could_ be functional rather than class-based. That'd be more similar to the mistral-finetune repo that Aldo pointed me toward.

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+    setup_logger(_args.log_level)
+
+    distributed_init(_args)
+
+    dataw = MultipackDataWrapper(
+        model_name_or_path=_args.model_name_or_path,
+        data_path=_args.data_path,
+        effective_batch_size=_args.effective_batch_size,
+        max_batch_len=_args.max_batch_len,
+        world_size= int(os.getenv("WORLD_SIZE")),
```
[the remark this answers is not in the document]
JamesKunstle: we initialize distributed a bit earlier, do you suggest swapping it out?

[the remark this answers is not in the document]
JamesKunstle: also, there isn't a torch.distribted.get_local_rank()- is it typical to just use

```
local_rank = torch.distributed.get_rank() // torch.distributed.get_world_size()
```
?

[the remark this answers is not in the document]
fabianlim: @JamesKunstle No that is not reliable. Actually `torch` added a new method `get_node_local_rank` but that is only available in bleeding edge https://github.com/pytorch/pytorch/pull/123992/files, so we are out of luck.  My suggestion is:
- follow the official implementation above.
- wrap this logic in function or patch it to `torch.distributed`  (the latter will be better to keep the api future proof)

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+        self.world_size: int = world_size
+        self.is_padding_free: bool = is_padding_free
+        self.seed: int = seed
+
+        self.dataset = setup_dataset(data_path=data_path)
+        self.tokenizer = setup_tokenizer(model_name_or_path)
+
+        self.packing_max_batch_len, self.grad_accum = (
+            find_packing_max_batch_len_and_grad_accum(
+                num_gpus=torch.distributed.get_world_size(),
```
fabianlim, line 66: we already have `world_size` above so better to just get it from there.

src/instructlab/training/trainer.py
```diff
@@ -0,0 +1,648 @@
...
+                effective_batch_size=effective_batch_size,
+                max_batch_len_per_gpu=max_batch_len,
+            )
+        )
+
+        self.samples_per_gpu = effective_batch_size // self.grad_accum // world_size
+        self.data_loader = setup_dataloader(
+            self.dataset,
+            self.tokenizer.pad_token_id,
+            num_workers=8,
```
fabianlim, line 77: maybe do not hardcode this, let this be an input to the function/wrapper

src/instructlab/training/trainer.py
fabianlim: yes i sugggest to do any `automodel` or `dolomite` loading outside of the dataset and deepspeed preperation functions, then just pass it in to those functions as needed.

src/instructlab/training/trainer.py
fabianlim: ok i misunderstood. the flag is `is_padding_free`. 

But bear in mind that it is typically not canon (at least in the huggingface world) that a dataloader is prepared for a trainer, usually a dataset is passed to a trainer. 

Aslo its not canon that a prepared model is passed to the trainer (at least in huggingface world). Usually a unprepared model (i.e. not yet wrapped with deepspeed) is passed to the trainer and it does it

Hence what im saying is that I feel you really only need one class `DeepSpeedTrainer`

JamesKunstle: ++ okay @fabianlim that makes a lot of sense.

JamesKunstle:
@fabianlim, @aldopareja and I are going to reorganize this effort. Based on both of your comments it seems like there's a more lightweight, canonical, and less-coupled way of doing this. I tried to minimize rewrite with this PR and slice stuff how it might be most reusable but it seems that there's a better way.

PR #23, opened 2024-06-15
Ap/black formatting

aldopareja:


russellb (changes requested):
if you do this, can you also introduce the automation you used to make sure it stays formatted according to whatever standards?

Ideally we'd use the same tools + configs as the other repos in instructlab though ...

PR #25, opened 2024-06-17
More Checkpointing Robustness

fabianlim:
- add universal checkpointing

To verify
1. train until a `ds_native` checkpoint is saved, stop the `torchrun` process.
2. restart `torchrun` with a different `num_proc` while using the *exact same train command*, 
3. observe printouts for creating the universal checkpoint in the `latest` directory. Training will resume with the different `num_proc`

Results on Dolomite full FT
- wrong learning rate was inadvertently used here, but it does prove that the method works.

image

Results on Dolomite QLoRA

image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ py-cpuinfo
 # replace custom pytorch images with the 2.3.0
 torch>=2.3.0a0
 transformers==4.41.2
-datasets
+datasets>=2.15.0
 numba
 numpy
 rich
@@ -22,5 +22,5 @@ pydantic>=2.7.0
 # 
 
 # deepspeed needs to be at the bottom or it'll break during installation
-deepspeed==0.14.3
+deepspeed>=0.14.3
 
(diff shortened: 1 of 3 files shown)


aldopareja:
@fabianlim, I'm approving, before merging can you include the set of tests you did with the matching plots?, that way we know what doesn't need any more testing.

aldopareja (approved):

aldopareja:
After including the tests feel free to merge.

aldopareja:
Fabian, I think we should make it optional to load from latest?. What if you want to load from a previous checkpoint?. Or what if you don't want to load at all and just start fresh?.

fabianlim:
> Fabian, I think we should make it optional to load from latest?. What if you want to load from a previous checkpoint?. Or what if you don't want to load at all and just start fresh?.

then you just start from a fresh `output_dir`

aldopareja:
sounds right then, let's merge this

aldopareja:
the linter is complaining, can you fix that?

russellb:
In the future, can you squash commits like "improvements" and "lint + format" into previous commits prior to merging? Thanks!

Issue #51, opened 2024-06-19
Datapoints larger than max_batch_len included in sampler output


ChanderG:
I was trying to run training on a single GPU (so small max batch lengths), when I got the following error:
```
File "/opt/conda/envs/ai/lib/python3.11/site-packages/dolomite_engine/hf_models/models/gpt_dolomite/main.py", line 116, in forward
    input_ids, position_ids, token_type_ids, labels, cu_seqlens, max_seqlen = self.prepare_inputs_for_model(
                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/ai/lib/python3.11/site-packages/dolomite_engine/hf_models/models/gpt_dolomite/base.py", line 89, in prepare_inputs_for_model
    check_list_type(input_ids, error_message.format(variable="input_ids", dtype="int"))
  File "/opt/conda/envs/ai/lib/python3.11/site-packages/dolomite_engine/hf_models/utils.py", line 11, in check_list_type
    assert isinstance(list_of_list[0], list), error_message
                      ~~~~~~~~~~~~^^^
IndexError: list index out of range
```

On investigating, I found that the reason was that the DataLoader was returning empty batches. The issue was that the Collation function filters out entries longer that `max_batch_len`, but the Multipack Sampler was still returning batches with samples longer than max_batch_len. This underlying issue doesn't manifiest in most scenarios, due to this extra processing.

Here is a minimal reproducible example:
```
  from multipack_sampler import MultipackDistributedBatchSampler
  import numpy as np

  lengths = np.array([1000, 1000, 2000, 2000, 2000, 4000])
  sampler = MultipackDistributedBatchSampler(batch_max_length=2500, lengths=lengths, num_replicas=1, rank=0, seed=42, padding=True)

  for i in sampler:
      print(i)
```

We expect "4000" to not show up when the batch_max_len is set to 2500, but the output is something like the following:
```
[3]
[2]
[5]
[4]
[0 1]
```
(where index 5 is included)

PS: may be good to have a test suite for the Multipack sampler.

PR #52, opened 2024-06-19
bugfix: remove out large samples from the multi pack batch sampler

ChanderG:
Fix for #51 

larger entries, longer than max_batch_len, may be present in the output of the sampler which gets filtered out in the collator

in rare instances, this can lead to empty batches which can be problematic for the training loop


aldopareja:
Great catch! This should be processed in the data processing step. Can you add a warning in red or yellow 033 notation so the user knows?

aldopareja:
You can just add it when there's a sample larger than Max batch lenght

aldopareja (approved):

PR #58, opened 2024-06-21
Ap/async logger

aldopareja:


RobotSail (approved):
LGTM

russellb:
@RobotSail @aldopareja, we really, REALLY can't merge PRs that fail the DCO check in CI. This is a legal requirement for how we accept contributions to the repos.

requirements.txt
```diff
@@ -7,7 +7,7 @@ torch>=2.3.0a0
 transformers>=4.41.2
 datasets>=2.15.0
 numba
-numpy
+numpy==1.26.4
```
russellb, line 10: specific version pinning like this is very problematic as this gets included in a larger distribution. This should be removed ASAP

RobotSail:
@russellb You are right about the DCO and version pinning, thank you for pointing it out.

PR #65, opened 2024-06-24
add missing deepspeed options

### LINKED ISSUE #57: Deepspeed CPU offloader isnt being passed
Right now if you enable cpu offloading it doesn't actually pass the `--cpu_offload_optimizer` flag or any of the other deepspeed specific options.

### LINKED ISSUE #62: Add separate "save_samples_ds" argument into the interface
Currently the only interface option is for regular hf format checkpointing. There needs to be an option for ds universal checkpoints for auto-resume

RobotSail:
Fixes #62 
Fixes #57 

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -90,6 +90,9 @@ class DeepSpeedOptions(BaseModel):
     cpu_offload_optimizer_ratio: float = 1
     cpu_offload_optimizer_pin_memory: bool = False
 
+    # don't save in deepspeed format as a default
+    save_samples: int | None = None
+
 
 class TrainingArgs(BaseModel):
     """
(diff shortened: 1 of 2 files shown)


aldopareja:
do we really want to expose that?, we might as well just ask the user to provide a json file with the deepspeed config they want.

aldopareja:
but I think in general exposing so much to the user is overwhelming, no one really changes these options anyway, and if you do, then you can also just hard code it in the code instead of using a parameter.

kdvalin:
For CPU Offload Memory Pinning, [this constructor](https://github.com/instructlab/training/blob/a9f97c8771d421afc085f123806467ca39ac0131/src/instructlab/training/main_ds.py#L243) needs `cpu_offload_optimizer_pin_memory=args.cpu_offload_optimizer_pin_memory,` added.  There seems to be other options missing from this as well.

RobotSail:
@aldopareja Yes this is something that should be configurable. In the future we can look at exposing the full config, but right now much of this stuff is conditional so we're constructing it ourselves.

Maxusmusti:
yeah, @aldopareja these options are actually important for us to expose for the performance team:
```
cpu_offload_optimizer
cpu_offload_optimizer_ratio
cpu_offload_optimizer_pin_memory
```
but agreed that we don't need to expose everything as it can get quickly overwhelming. I think the current state of this PR is good

Maxusmusti (approved):
LGTM

PR #66, opened 2024-06-25
convert enum to str when referencing

cdoern:
click does not support an enum type for click.choice. This means that in the CLI we need to somehow get from str -> enum to pass to the library.

Adding this use_enum_values, allows a user to specify things like `nf4` and `None` as strs in their config.yaml and they are seen as valid by the library.


JamesKunstle:
Understood once I read the docs:

use_enum_value: 
> Whether to populate models with the value property of enums, rather than the raw enum. This may be useful if you want to serialize model.model_dump() later. Defaults to False.

JamesKunstle (approved):

RobotSail (approved):
LGTM

PR #68, opened 2024-06-25
requirements: Adjust numpy dependency range

russellb:
According to comments on #61, numpy 2 doesn't work with this library
right now. Instead of pinning to a specific 1.x version, allow a range
from a known working version up to, but not including 2.0.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,13 @@ torch>=2.3.0a0
 transformers>=4.41.2
 datasets>=2.15.0
 numba
-numpy==1.26.4
+# Note: numpy ranges copied from instructlab/instructlab
+#
+# HabanaLabs / Intel Gaudi env comes with Python 3.10 and slightly older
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -7,7 +7,7 @@ torch>=2.3.0a0
 transformers>=4.41.2
 datasets>=2.15.0
 numba
-numpy==1.26.4
+numpy>=1.26.4,<2.0.0
```
RobotSail, line 10: Please change this line to:
```suggestion
numpy<2
```

I believe the `1.26.4` was arbitrarily pinned, it just needs to be under 2.

RobotSail (commented):
Just need to change how we're pinning the version but otherwise LGTM

requirements.txt
russellb: got it.  I'm actually going to copy over what the CLI repo has, because it had some more specific reasons on versions

aldopareja (approved):
lgtm

PR #70, opened 2024-06-25
Fix broken CI

russellb:
Past PRs were merged that broke CI jobs. Ideally this should never be
happening. Get the job back to a passing state.

### DIFF
--- a/src/instructlab/training/tokenizer_utils.py
+++ b/src/instructlab/training/tokenizer_utils.py
@@ -10,11 +10,11 @@ from instructlab.training.utils import log_rank_0
 
 @dataclass
 class SpecialTokens:
-    system: str = field(default=None)
+    system: str = field(default="")
     user: str = field(default="<|user|>")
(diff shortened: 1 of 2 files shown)


src/instructlab/training/tokenizer_utils.py
```diff
@@ -10,11 +10,11 @@
 
 @dataclass
 class SpecialTokens:
-    system: str = field(default=None)
+    system: str = field(default="")
```
aldopareja, line 13: we are checking for `is none` somewhere so it scares me that this would break it.

aldopareja (approved):

src/instructlab/training/tokenizer_utils.py
russellb: It failed the `mypy` check because the type is declared as a string, but the default type is not a string

russellb: the alternative is to fix the type declaration

PR #74, opened 2024-06-25
remove e2e for now

### LINKED ISSUE #73: Remove e2e test
The end-to-end test introduced in PR #63 uses `ilab` to run our code. This is wrong because the training library doesn't rely on the `ilab` CLI.

RobotSail:
fixes #73 

We're removing the end-to-end test because these rely on the implementation of `ilab`. 
What we want instead is to have e2e tests that are based *solely* on the code of this repository, and not a separate tool consuming `instructlab-training` internally. 


Maxusmusti (approved):
LGTM

nathan-weinberg:
Why was spellcheck removed here? That has nothing to do with E2E

PR #76, opened 2024-06-25
Remove e2e 2 the electric boogaloo

RobotSail:
We're removing the end-to-end test because these rely on the implementation of ilab.
What we want instead is to have e2e tests that are based solely on the code of this repository, and not a separate tool consuming instructlab-training internally.

Fixes #74


Maxusmusti (approved):
LGTM

nathan-weinberg:
Why was spellcheck removed here? That has nothing to do with E2E

russellb:
> We're removing the end-to-end test because these rely on the implementation of ilab. What we want instead is to have e2e tests that are based solely on the code of this repository, and not a separate tool consuming instructlab-training internally.
> 
> Fixes #74

Testing the user end-to-end instructlab workflow is the point of this workflow. It will give you an early indication if your changes to the library break the CLI. Sometimes that might be on purpose, but in case it's not, it will let you know before you publish a release.

This is not intended to replace other functional testing aimed at the library more directly, but the full end-to-end workflow is important.

I'm OK if you want to disable it for the moment, but it needs to go back on at some point.  A better way to do that is to just edit the workflow to turn it off instead of removing it from the repo.

Issue #63, opened 2024-06-24
Add end-to-end automated test workflow


russellb:
Once this library is being consumed by `instructlab/instructlab`, it would be helpful to automatically test changes to this library in an end-to-end run of the `ilab` workflow. You can copy this workflow that I added to the sdg repo to do this. It should work with no modifications, though it does say "instructlab-sdg" in one comment.

https://github.com/instructlab/sdg/pull/33

russellb:
This was reverted in #76. My comment there explains why this is relevant:

> Testing the user end-to-end instructlab workflow is the point of this workflow. It will give you an early indication if your changes to the library break the CLI. Sometimes that might be on purpose, but in case it's not, it will let you know before you publish a release.
> 
> This is not intended to replace other functional testing aimed at the library more directly, but the full end-to-end workflow is important.
> 
> I'm OK if you want to disable it for the moment, but it needs to go back on at some point. A better way to do that is to just edit the workflow to turn it off instead of removing it from the repo.

PR #31, opened 2024-06-17
Implement initial CI configuration

nathan-weinberg:
Supersedes #7 

This PR adds the following CI jobs:
- Dependabot
- GitHub Action linting
- Markdown Linting (with fixes)
- Spellcheck (with fixes)
- Code Linting (with **some** fixes - see note below)

As well as a `Makefile` for local running of the checks.

**Additional Notes**
- I will do a followup for PyPI packaging but I want to work directly with the folks from Training Squad on that to keep them in the loop
- WRT the Python checks - @RobotSail @Maxusmusti @JamesKunstle I didn't go through all the failures (the fixes I did were auto-format fixes that have no functional impact) because I don't want to hamper y'alls velocity and break things currently working - so the checks remain but are non-blocking for now - I would encourage y'all to address these when it makes sense

PR #77, opened 2024-06-26
Revert "remove incorrect end-to-end implementation"

russellb:
This reverts commit ebc5e317a1269b64a5b1586c8b07364490d59add from #76.

Here is my comment on #76 that better explains how this workflow fits
in to the bigger picture:

> Testing the user end-to-end instructlab workflow is the point of
> this workflow. It will give you an early indication if your changes to
> the library break the CLI. Sometimes that might be on purpose, but in
> case it's not, it will let you know before you publish a release.
>
> This is not intended to replace other functional testing aimed at the
> library more directly, but the full end-to-end workflow is important.
>
> I'm OK if you want to disable it for the moment, but it needs to go
> back on at some point. A better way to do that is to just edit the
> workflow to turn it off instead of removing it from the repo.

Issue #63

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E test
+
+on:
+  push:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.py'
+      - 'pyproject.toml'
+      - 'requirements*.txt'
+      - '.github/workflows/e2e.yml'
+  pull_request:
+    branches:
+      - "main"
+      - "release-**"
+    paths:
+      - '**.py'
+      - 'pyproject.toml'
+      - 'requirements*.txt'
+      - '.github/workflows/e2e.yml'
+
+jobs:
+  e2e:
+    runs-on: ubuntu-gpu
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      # No step-security/harden-runner since this is a self-hosted runner
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          # https://github.com/actions/checkout/issues/249
+          fetch-depth: 0
+
+      - name: Install Packages
+        run: |
+          sudo apt-get install -y cuda-toolkit git cmake build-essential virtualenv
+          nvidia-smi
+          sudo ls -l /dev/nvidia*
+
+      - name: Setup Python 3.11
+        uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+        with:
+          python-version: 3.11
+          cache: pip
+          cache-dependency-path: |
+            **/pyproject.toml
+            **/requirements*.txt
+
+      - name: Remove llama-cpp-python from cache
+        run: |
+          pip cache remove llama_cpp_python
+
+      - name: Cache huggingface
+        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+        with:
+          path: ~/.cache/huggingface
+          # config contains DEFAULT_MODEL
+          key: huggingface-${{ hashFiles('src/instructlab/configuration.py') }}
+
+      - name: Install instructlab and libraries
+        run: |
+          export PATH="/home/runner/.local/bin:/usr/local/cuda/bin:$PATH"
+          python3 -m venv venv
+          . venv/bin/activate
+          git clone https://github.com/instructlab/instructlab.git
+          cd instructlab
+          sed 's/\[.*\]//' requirements.txt > constraints.txt
+          python3 -m pip cache remove llama_cpp_python
+          CMAKE_ARGS="-DLLAMA_CUBLAS=on" python3 -m pip install --no-binary llama_cpp_python -c constraints.txt llama_cpp_python
+          # needed for --4-bit-quant option to ilab train
+          python3 -m pip install bitsandbytes
+          # install instructlab
+          python3 -m pip install .
+          cd ..
+          # Install instructlab-sdg
+          python3 -m pip install .
+
+      - name: Run e2e test
+        run: |
+          . venv/bin/activate
+          ./instructlab/scripts/basic-workflow-tests.sh -cm
+
+      - name: Remove llama-cpp-python from cache
+        if: always()
+        run: |
+          pip cache remove llama_cpp_python


russellb:
You don't have to merge this right away if it was causing a short-term problem, but I'm staging it here for your convenience.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
```
RobotSail, line 1: Remove the spellchecker please.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+          # install instructlab
+          python3 -m pip install .
+          cd ..
+          # Install instructlab-sdg
+          python3 -m pip install .
+
+      - name: Run e2e test
+        run: |
+          . venv/bin/activate
+          ./instructlab/scripts/basic-workflow-tests.sh -cm
```
RobotSail, line 87: Why are we running tests that don't live in this repo? We want all of our end-to-end tests to be ones that live inside of the training repo.

RobotSail (changes requested):
Some thoughts:

1. We want the end-to-end tests to be structured in a way where we have a clean file consuming the latest version of our training library. The tests should live in this repo so that they can be adjusted, and the proper advisories can be made WRT updating the upstream CLI repo.
2. We are not planning on using the spell-checker going forward.

.github/workflows/spellcheck.yml
russellb: your PR made that look like an accident since it wasn't mentioned. I don't care either way

.github/workflows/e2e.yml
russellb: did you read my explanation?  maybe we just need to talk live

n1hility:
So while we might have chosen to break things into separate repos, I don't think we agreed to run them as completely independent projects with their own processes and governance. We still need to share and build off of each other IMO.

.github/workflows/e2e.yml
RobotSail: @russellb Are these end-to-end tests the biggest priority currently?

.github/workflows/spellcheck.yml
RobotSail: Yes we intend to drop the spell-checker going forward. We squash the commits so it might have been dropped somehow.

.github/workflows/e2e.yml
russellb: testing that everything works together when run via the end-user CLI workflow?  yes

RobotSail:
@n1hility I don't disagree, and would like to introduce e2e tests that don't break consuming libraries. The issue with this implementation is that these tests live beyond our repo, and therefore are very fragile. We would like to keep everything within the same repo. Happy to implement this myself after the 15th.

russellb:
Relevant note: this test script has not been updated to use the new training code.  It was changed to use legacy mode when new training was merged into the CLI.  At the moment, the test coverage would be ensuring that everything can still be installed together, including whatever changes are present in a training PR.

Once this issue is resolved, it would also ensure that `ilab model train` works, at least in some basic configuration.

https://github.com/instructlab/instructlab/issues/1470

n1hility:
> @n1hility I don't disagree, and would like to introduce e2e tests that don't break consuming libraries. The issue with this implementation is that these tests live beyond our repo, and therefore are very fragile. We would like to keep everything within the same repo. Happy to implement this myself after the 15th.

Can you elaborate how it's fragile? This is a pretty common practice for e2e testing the combination of a library into a primary consumer, you run a smoke test of the consumer in your project so that you catch compatibility problems.

RobotSail:
> Can you elaborate how it's fragile? 

Having end-to-end tests in a separate repo means that if we make a breaking change, we would need to fix the tests in the upstream to get them working on our end. It makes it so we are coupled to the upstream repo, and if these tests are to break for any reason, this would impact our PR status downstream.

On the other hand, defining an end-to-end test suite within this repo allows us to ensure that it's reflective of how we intend this library to be used. This also makes it much more straightforward to develop the tests as we build out the library.

In either case we'd be able to detect breakages, but one allows us much more freedom with how we go about testing the library.

.github/workflows/spellcheck.yml
nathan-weinberg: There is no need to remove the spellchecker. It is a standard action that is used throughout the entire organization.

However, if you have a legitimate reason for removal, I'm happy to hear you out on it!

.github/workflows/e2e.yml
nathan-weinberg: @n1hility has expressed concern around interdependancy/functional regressions between the various components, so it seems important to have this type of test

nathan-weinberg (dismissed):

.github/workflows/spellcheck.yml
nathan-weinberg: I will point out that yourself and three other maintainers approved the implementation: https://github.com/instructlab/training/pull/31

But again, if something has changed, I'm willing to hear it out. I think an explanation is warranted though. as @russellb this removal wasn't explained in any PR description or commit message.

nathan-weinberg:
I'll just add to the debate of the value of the E2E that we've added it to the Eval repo and it has already helped identify the need for a dependency bump that @danmcp implemented: https://github.com/instructlab/eval/pull/14#issuecomment-2189981189

n1hility:
> > Can you elaborate how it's fragile?
> 
> Having end-to-end tests in a separate repo means that if we make a breaking change, we would need to fix the tests in the upstream to get them working on our end. It makes it so we are coupled to the upstream repo, and if these tests are to break for any reason, this would impact our PR status downstream.
> 
> On the other hand, defining an end-to-end test suite within this repo allows us to ensure that it's reflective of how we intend this library to be used. This also makes it much more straightforward to develop the tests as we build out the library.
> 
> In either case we'd be able to detect breakages, but one allows us much more freedom with how we go about testing the library.

Thanks for sharing your thinking. What you describe is really mock integration testing than a form e2e testing. Both have their uses and are complementary, but mock testing is trading real world accuracy for speed, since you are only testing what you anticipate the interaction to be vs what it actually is. This makes a lot of sense when e2e is too slow for PR testing, since better to run it nightly and unit + mock tests in PR. However in this case, it's a fast test suite (only 30 mins mostly from build), and the alternative we have right now is no testing. Granted this is only testing the legacy training path, so its value is more limited than what it could be. However it will catch API breakage and dependency issues and some interaction. 

While a planned API change will cause these to fail, it's pretty easy to suspend them, or alternatively sustain compatibility until you update the dependent ilab project.

RobotSail:
A few things:

- 30 mins is *way* too slow for an end-to-end test. At the pace we push PRs, these will be reset often and will likely go to waste.
- The current e2e test which was added still failed to catch a bug as it showed everything passed successfully.
- Presently, we enforce our library's interface via pydantic, so everything in terms of options will either default or throw an error. Beyond that, once `run_training` is invoked would be that which lives in this repo. Meaning that both end-to-end tests would be checking the same functionality.

.github/workflows/spellcheck.yml
RobotSail: We don't feel that a spell-checker is a necessary check that we need to run on PRs. Furthermore, if typos exist and community members spot them, then that becomes another window for people to make easy contributions.

nathan-weinberg: I'm not really following here - it doesn't sound like anything has changed since the original approval? I'm also not quite sure who the "we" is you are referring to - I haven't seen any engagement on this topic apart from yourself, perhaps if others feel the same they can chime into the discussion? 

As for your second point, that is also not quite clear to me - why would we have humans manually reading documents for spelling errors when this is something that is already automated in a standardized way across this entire org?

lhawthorn: Speaking as a community engagement human who knows how much it means to someone to contribute, even if it's just to fix a silly spelling error, I feel like spell checking is unnecessary. Will we have misspelled words? Yes, we will. Will it give us a chance to have low hanging fruit work items for people to address? Yes, it will. And it's those simple fixes that get people started doing bigger things. My .02 EUR, strong opinion, fairly strongly held even.

cdoern:
> 30 mins is way too slow for an end-to-end test. At the pace we push PRs, these will be reset often and will likely go to waste.

This is a pretty standard (if not fast) time for an e2e, I have seen ones in OCP that take hours. 

Sorry to jump into the convo super late here, but in all honesty I think what this entire conversations boils down to is:

speed of development vs quality of code

We have been struggling to find a balance here. So this is what I would suggest. I think we should add back the e2e, but with the full understanding that it might break sometimes with immediate or pretty quick fixes. As long as we can use the test to pinpoint which changes introduced an issue, that is better than no e2e AND it is better than stalling necessary development.

We need to find a middle ground here, adding the test but understanding the circumstances seems pretty measured to me.

I understand both perspectives here and I am open to any feedback on this stance!

.github/workflows/spellcheck.yml
Maxusmusti: > who the "we" is
@nathan-weinberg I can add here that most of our current contributors (IBM India/Singapore or Cambridge, Red Hat, etc.) have aired grievances about the check. This isn't really @RobotSail's fault as these were messages in Red Hat / IBM slack groups. I will make sure that going forward, if any contributors have concerns that they create a public GH issue so that the conversation can be more accessible, that's my bad for not insisting upon it originally.

nathan-weinberg: A valid perspective, but something else to consider is that this was removed without community input. I think there's a conversation to be had here, but given the nature of how this action was removed, IMO the following should be done:
1. It is readded in this PR, returning us to the status quo.
2. We have the conversation, understanding the context that this is something that should be consistent org-wide, whether the spellcheck is present or not.
3. If the conversation goes the way of removing it, it is done then properly in a followup, without being piggybacked onto the E2E (as was the case when it was removed).

This I think holds up to the standards of the open source governance of this project. I think having something removed that was previously approved and having to justify the returning to the status quo does not follow these standards and is not a precedent we want to set. What do you think @lhawthorn?

nathan-weinberg: @Maxusmusti Not at all! I appreciate this perspective being shared, it's important context that helps shape the conversation!

russellb: I've dropped spellcheck from this PR. It's really not important at all compared to e2e testing. We can discuss elsewhere later.

russellb:
Updated the PR to drop re-adding spellcheck. It was included in the commit I was reverting, but is not in scope for what I really care about here.

RobotSail:
So after a thorough discussion among the training team, our decision is as follows:
- It's OK if these tests live in the instructlab repo
- The end-to-end tests must use _our_ training library in the case of training
- Optimizations should be made so that the end-to-end test takes the minimum amount of time needed to train. This shouldn't be a 30 minute process if at all possible.

We feel that this is only fair considering these tests live beyond our repository.

RobotSail:
After further discussion with related parties, we came to the conclusion that these e2e tests will be going in.

RobotSail:
This PR can be merged after https://github.com/instructlab/instructlab/pull/1494

booxter:
@RobotSail can this merge now?

RobotSail:
@russellb Can could you please update this test so that it's running with the full training flag, e.g. `-fT`? 

https://github.com/instructlab/instructlab/blob/323e503002ca314c7b44ec46fd5b00318a960cee/scripts/basic-workflow-tests.sh#L480

russellb:
This has gone stale at this point. 

I would suggest considering an EC2-based workflow instead of this that ran on a github runner. That might as well be a new PR.

PR #78, opened 2024-06-26
Another Fix for Checkpointing

fabianlim:
It will be safer to only allow a single rank to remove the directory because there is a possibility of race condition


aldopareja (approved):

russellb:
@fabianlim the PR description has an explanation of the change, but the commit message only says "bugfix". In the future, it would be great if you could include the same detail in the commit log, as well. Thanks!

fabianlim:
@russellb got it! Sorry for any inconvenience caused!

PR #79, opened 2024-06-26
Reduce Unnecessary Multiprocessing

Maxusmusti:
We are spawning more threads than needed during data process, and using multiprocessing to get data lengths in a scenario where a single list comprehension is in almost all cases faster.


aldopareja (approved):
Lgtm

PR #80, opened 2024-06-27
Add LR scheduler and Fix Granite Checkpoint Loading

fabianlim:
This PR adds the LR scheduler configuration and also fixes the granite loading issue for multinode settings.
- this PR subsumes the changes in @aldo-pareja 's `lr_scheduler` branch, so a seperate merge of that would not be necessary
- the strategy now for converting the granite checkpoint, is that each node will elect one local rank to do the convert.
- the conversion will be one on the `args.output_dir`. There is no assumption it is the same for all nodes or not.  But assuming they might be the same, the `tmpdir` is shared by a `.{group_rank}` postfix. See [group rank](https://pytorch.org/docs/stable/elastic/run.html#environment-variables)
- tested on multi node

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -231,7 +231,7 @@ def setup_model(args, tokenizer, train_loader, grad_accum):
         )
 
     lr_scheduler = get_scheduler(
-        name="cosine",
+        name=args.lr_scheduler,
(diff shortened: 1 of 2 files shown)


Maxusmusti (approved):
LGTM, tested on single-node as well and worked. @aldo-pareja have you tested on multi-node?

PR #112, opened 2024-06-28
Add Control To Turn Off Multipack

fabianlim:
Multipack can sometimes cause a GPU to be assigned 0 samples if the effective batch size is very small. 
- this provides an additional control `sampler`, and can be set to `"distributed"` or `"multipack"` (default)
- The `distributed` version is just a random sampler. However, we will still call `find_packing_max_batch_len_and_grad_accum` to get the `grad_accum` and `samples_per_gpu`


aldopareja (approved):
lgtm :)

PR #114, opened 2024-06-28
Catch Error if Packing Max Batch Len Cannot be Found

fabianlim:
this will remove the control for sampler
- instead sampler will be decided if the max batch len cannot be found.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -508,7 +508,7 @@ def main(args):
         args.sampler = "multipack"
     except RuntimeError as e:
         if os.environ["LOCAL_RANK"] == "0":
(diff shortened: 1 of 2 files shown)


src/instructlab/training/main_ds.py
```diff
@@ -506,7 +506,10 @@ def main(args):
             seed=args.seed,
         )
         args.sampler = "multipack"
-    except RuntimeError:
+    except RuntimeError as e:
+        if os.environ["LOCAL_RANK"] == "0":
+            print (f"\033[38;5;120m{e}\033[0m")
```
Maxusmusti, line 511:
```suggestion
print(f"\033[38;5;120m{e}\033[0m")
```

PR #116, opened 2024-06-30
Add lazy import for `run_training`

tiran:
The `main_ds` module imports DeepSpeed and PyTorch. These imports take a very long time and slow down startup of the Instruct CLI `ilab` tool by over a second.

Use PEP 562 hooks to implement lazy import and loading for `run_training` function. The function is imported when the attribute is accessed.

See: https://github.com/instructlab/instructlab/issues/1467


RobotSail (approved):
Thank you for this change, LGTM!

Issue #92, opened 2024-06-26
FlashAttention disable support


ktam3:

Maxusmusti:
In-progress resolution: #15

Maxusmusti:
PR #118 merged to add functionality

PR #118, opened 2024-07-01
Flash Attention Disable Toggle (Take 2)

Maxusmusti:
For hardware scenarios where flash attention is potentially not supported.

(Re-implementation of now out-of-date #15)

For issue #92

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -4,6 +4,7 @@ Collection of config objects used in the InstructLab training library.
 
 # Standard
 from enum import Enum
+from typing import Optional
 import os
 
 # Third Party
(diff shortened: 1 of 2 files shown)


Maxusmusti:
Tested all four options:
 - disable_flash_attn=False, is_padding_free=False
 - disable_flash_attn=True, is_padding_free=False
 - disable_flash_attn=False, is_padding_free=True
 - disable_flash_attn=True, is_padding_free=True

src/instructlab/training/config.py
```diff
@@ -143,6 +143,8 @@ class TrainingArgs(BaseModel):
         )
     )
 
+    disable_flash_attn: bool = False
```
cdoern, line 146: can we make this explicitly `Optional[bool]`. I think we will hit errors otherwise at times

Maxusmusti: Ok, can do. Should we also do the same for the `mock_data` field in the config? Because that operates the same way

aldopareja (approved):

Maxusmusti:
Tested and validated on V100s

PR #119, opened 2024-07-01
added data processing logging

aldopareja:


src/instructlab/training/data_process.py
```diff
@@ -193,6 +194,7 @@ def remove_pretrain_system_messages(example: dict):
 
 
 def main(args: DataProcessArgs):
+    print(json.dumps(args.__dict__))
```
JamesKunstle, line 197: do we have a logger we could use for this instead of printing directly?

JamesKunstle (approved):
lgtm, one question to consider but it can go in

src/instructlab/training/data_process.py
aldopareja: we could have a logger but it's used mostly on the training repo.

JamesKunstle:
@aldo-pareja this can go in if you fix linting/DCO/rebase.

RobotSail (approved):
LGTM

RobotSail:
@aldo-pareja It looks like we added this in a separate PR when we converted this script into a callable SDK.

PR #121, opened 2024-07-02
Use deferred imports instead of lazy loading

tiran:
PR #116 added lazy loading of `run_training` function with a PEP 562 `__getattr__` module hook. Pylint does not understand the trick, which caused some problems. Replace the lazy hook with a wrapper function and deferred imports.

Also mark all public APIs with a comment, so developers are aware which classes are used by other packages.


RobotSail (approved):
LGTM

nathan-weinberg (approved):

PR #122, opened 2024-07-02
🐛: fix optimizer selection logic so that FusedAdam is never loaded when CPU offloading is enabled

RobotSail:
Basically the title. Just changes the logic so we dont load FusedAdam unnecessarily.


Maxusmusti (approved):
LGTM

Issue #136, opened 2024-07-15
Update Data Processing and Sampling Selection


Maxusmusti:
Data Processing must support pretraining with padded special tokens

Sampler auto-select must work with these pretrain samples (supposedly isn't right now)

bbrowning:
Does #137 fully resolve this issue, or is there another follow-up PR needed to take care of getting the sampler auto-select to work with the pretrain samples?

Maxusmusti:
@bbrowning this is fully resolved, the issue IBMers were running into turned out to be due to them experimenting on an older branch that was made before auto-selection. Thanks for checking! I had forgot to close this issue

PR #137, opened 2024-07-15
Updating token masking to support pretraining w/ masked special tokens

Maxusmusti:
For issue #136

### DIFF
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -1,9 +1,9 @@
 # Standard
 from functools import partial
 from pathlib import Path
-from typing import List
 import logging
 import os
+import random
 
 # Third Party
 from datasets import load_dataset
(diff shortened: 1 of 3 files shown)


src/instructlab/training/data_process.py
```diff
@@ -40,22 +41,12 @@ def check_valid_sample(
     eos_token_index = (whole_sentence_tk == eos_tk).nonzero()[0]
 
     # check that there are at least one user_token, assistant_token and eos_token
```
aldopareja, line 43: remove this comment.

src/instructlab/training/data_process.py
```diff
@@ -64,123 +55,134 @@ def check_valid_sample(
...
+    instruct_indices = []
+    for i in tqdm(range(len(data)), desc="Data type sorting"):
+        if data[i][0] == pretrain_tk:
+            pretrain_indices.append(i)
+        else:
+            instruct_indices.append(i)
+    return pretrain_indices, instruct_indices
 
 
 def remove_pretrain_system_messages(example: dict):
```
aldopareja, line 188: I think we should not remove pretrain system messages anymore.

aldopareja (approved):
just some small comments and then let's get it merged.

PR #143, opened 2024-07-17
fix: changes the check to check against both the enum option and enum value. Currently when we compare the type that gets passed in from the training args by only doing a check against the enum

RobotSail:
We update the check for qunatization to now check against both the enum option and enum option value.


Maxusmusti (approved):
LGTM

PR #151, opened 2024-07-24
Raise errors from run_training

booxter:


booxter:
Completely untested, just spitballing since I noticed that training failure doesn't break workflow tests for some reason.

src/instructlab/training/main_ds.py
```diff
@@ -674,17 +673,21 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
...
+        raise
+    else:
         print("\033[91mTerminating process 🤖\033[0m")
         process.terminate()
         try:
-            process.wait(timeout=60)
-        except subprocess.TimeoutExpired:
+            rc = process.wait(timeout=60)
+            if rc:
+                print(f"Process exited with code {rc}")
```
booxter, line 683: I see `print()`s here are using some color coding. Maybe I should do it here too. (But I don't like how color codes are embedded in each message instead of injected by a print wrapper... Perhaps this can also be improved.)

PR #156, opened 2024-08-02
Change success message at end of training

### LINKED ISSUE #155: Change success message for end of training
Currently the training loop prints out "termintating process" in red for both success and failure, let's fix it so that success prints out happy text in green. This way users aren't worried about a failure.

alinaryan:
Changes the verbiage and color of the success message to green. 

Fixes #155

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -677,8 +677,11 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
     finally:
         if "process" not in locals() or process is None:
             return
+        if process.poll() == 0:
+            print("\033[92mOperation completed successfully! 🎉\033[0m")
+        else:
+            print("\033[91mOperation failed, terminating process.\033[0m")
 
-        print("\033[91mTerminating process 🤖\033[0m")
         process.terminate()
         try:
             process.wait(timeout=60)


RobotSail (approved):
LGTM

aakankshaduggal (approved):

src/instructlab/training/main_ds.py
```diff
@@ -678,7 +678,7 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         if "process" not in locals() or process is None:
             return
 
-        print("\033[91mTerminating process 🤖\033[0m")
```
RobotSail, line 681: Will this be only for successful completions or all completions? Wondering whether we should somehow check the success status here.

RobotSail: It looks like you should be able to achieve this by using the return value from `process.poll()`:

```py
if process.poll() == 0:
  print("success!")
else:
  print("failed :(")
```

alinaryan: Thanks for the suggestion, Oleg! How's this?

RobotSail: It's perfect, thank you!

RobotSail (approved):
LGTM

cdoern (commented):
I am fine with this merging as long as we are 100% sure it will not introduce unexpected success/failed msgs. Is this something we want to risk for GA? I might be over-exaggerating here though since this seems pretty straight forward

RobotSail:
@cdoern This should be fine but we can test it in our e2e

Maxusmusti:
Pending test from Oleg, will approve/merge this PR once all looks good there. Will include this in a 0.4.0 release alongside James' PR once tested (#160)

RobotSail:
Tested this locally and it worked flawlessly. Thank you again for this great PR @alinaryan !

PR #160, opened 2024-08-06
Adds a flag to save checkpoints at the end of an epoch

JamesKunstle:
Currently, we save checkpoints:

1. Whenever we pass enough samples, and
2. sometimes at the end of training.

This adds saving per-epoch, so one could set save_samples really high and ONLY
save at the boundary of epochs.

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -139,6 +139,7 @@ class TrainingArgs(BaseModel):
     warmup_steps: int
     is_padding_free: bool
     random_seed: int = 42
+    checkpoint_at_epoch: bool = False
 
     mock_data: Optional[bool] = False
     mock_data_len: int = 0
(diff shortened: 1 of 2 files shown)


src/instructlab/training/main_ds.py
```diff
@@ -463,6 +463,15 @@ def train(args, model, tokenizer, train_loader, grad_accum, metric_logger):
             if local_rank == 0:
                 inner_pb.update(1)
             torch.cuda.empty_cache()
+
+        if args.checkpoint_at_epoch:
```
JamesKunstle, line 467: todo: check if `args.something` is `None` if `something` switch wasn't set

src/instructlab/training/config.py
```diff
@@ -139,6 +139,7 @@ class TrainingArgs(BaseModel):
     warmup_steps: int
     is_padding_free: bool
     random_seed: int = 42
+    checkpoint_at_epoch: bool = True
```
Maxusmusti, line 142: Make default `False` here, so as not to change default behavior (can make default true for phased training in CLI, but should probably avoid changing default before more proper epoch saving vs save samples update)

JamesKunstle: done

RobotSail (approved):
LGTM

JamesKunstle:
@Maxusmusti I just ran the code- the update is passed along fine.

PR #162, opened 2024-08-08
Accept tuples for lora.target_modules

booxter:
Pydantic field is defined as `list`, but in pydantic world, it means more than just lists. E.g. it allows tuples.

https://docs.pydantic.dev/2.0/usage/types/list_types/

This patch fixes the TypeError raised from the attempt to concatenate a list with a tuple.

Related: https://github.com/instructlab/instructlab/issues/2021


RobotSail (approved):
LGTM, it works

PR #168, opened 2024-08-14
Provide safeguards during training

RobotSail:
This PR adds safeguards to prevent the following issues:

- preventing the multipack sampler from being used when it cannot adequately process all samples
- prevent empty datasets from being provided or created
- prevent `max_batch_len` from being smaller than `max_seq_len`

### DIFF
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -204,9 +204,18 @@ def main(args: DataProcessArgs):
         {"additional_special_tokens": ["<|pretrain|>", "<|/pretrain|>"]}
     )
 
-    data = load_dataset("json", data_files=args.data_path, split="train")
+    try:
+        data = load_dataset("json", data_files=args.data_path, split="train")
+    except:
+        # pylint: disable=raise-missing-from,broad-exception-raised
+        raise Exception(
+            "Malformed or missing data, please ensure that your dataset is not empty and correctly formatted"
+        )
+
     if data.num_rows == 0:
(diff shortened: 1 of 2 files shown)


Maxusmusti (approved):
LGTM

src/instructlab/training/data_process.py
```diff
@@ -205,6 +205,10 @@ def main(args: DataProcessArgs):
     )
 
     data = load_dataset("json", data_files=args.data_path, split="train")
+    if data.num_rows == 0:
+        raise ValueError(
+            "The provided dataset is empty, please make sure that you were able to generate a dataset and try again."
```
JamesKunstle, line 210: I'd make this value error generation-agnostic.

src/instructlab/training/main_ds.py
```diff
@@ -555,6 +555,25 @@ def main(args):
         sampler=args.sampler,
         seed=args.seed,
     )
+    if len(train_loader) == 0:
```
JamesKunstle, line 558: this makes training slower so maybe we shouldn't automatically switch over to the distributed data sampler.

JamesKunstle: @Maxusmusti and I talked about it- defaulting to naive distribution seems reasonable

src/instructlab/training/data_process.py
RobotSail: Good call, changing this

Maxusmusti (approved):
Tested myself as well

PR #169, opened 2024-08-14
Ap/fix multipack for non granite models

aldopareja:
greatly increased efficiency by parallelizing many for loops, and also fixed the multipack sampler for non-granite models.

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -48,7 +48,10 @@ class DataProcessArgs(BaseModel):
     max_seq_len: int  # defines the max sequence length of a sample
     model_path: str  # either a HF model name or path to HF model
     chat_tmpl_path: str
-    num_parallel_procs: int
+    num_cpu_procs: int = Field(
+        default=16,
+        description="this is the number of CPU procs we use for data processing parallelization",
+    )
 
     # disable the protected namespace for the model_config field
     model_config = ConfigDict(protected_namespaces=())
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -48,6 +48,7 @@ class DataProcessArgs(BaseModel):
     max_seq_len: int  # defines the max sequence length of a sample
     model_path: str  # either a HF model name or path to HF model
     chat_tmpl_path: str
+    num_parallel_procs: int
 
     # disable the protected namespace for the model_config field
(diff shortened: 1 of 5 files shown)


src/instructlab/training/data_process.py
```diff
@@ -250,16 +242,22 @@ def main(args: DataProcessArgs):
...
     )
     log_rank_0(
         f"\033[33mnumber of dropped samples: {len(data) - len(data_with_input_ids)} -- out of {len(data)}\033[0m"
     )
 
     print("\033[92mCategorizing training data type...\033[0m")
-    pretrain_indices, instruct_indices = form_data_pools(
-        data_with_input_ids["input_ids"],
-        pretrain_tk=get_sp_token(tokenizer, "<|pretrain|>"),
+    # pretrain_indices, instruct_indices = form_data_pools(
```
Maxusmusti, line 252: NOTE: I can remove extraneous comments here

src/instructlab/training/config.py
```diff
@@ -48,6 +48,7 @@ class DataProcessArgs(BaseModel):
     max_seq_len: int  # defines the max sequence length of a sample
     model_path: str  # either a HF model name or path to HF model
     chat_tmpl_path: str
+    num_parallel_procs: int
```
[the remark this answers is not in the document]
RobotSail: It would also be good to document what this is for by doing:

```suggestion
    num_parallel_procs: int = Field(description="this is the number of CPUs we use for parallelization")
```

Also - is this a setting we could just automate without the user having to specify the number? I feel like that would be more ideal.

src/instructlab/training/token_dataset.py
```diff
@@ -17,10 +16,10 @@ class TokenDataset(Dataset):
     def __init__(self, data_path):
         self.data = load_dataset("json", data_files=data_path, split="train")
         self.lengths = np.array(
-            [
-                len(x["input_ids"])
-                for x in tqdm(self.data, desc="Data length calculation", colour="cyan")
-            ]
+            self.data.map(
```
Maxusmusti, line 19: NOTE: @aldopareja this is something we might want to discuss, as the switch to list comprehension resulted in notable speed-ups, so we may not want to switch back

Maxusmusti (approved):
LGTM, packing seems to work as expected

PR #174, opened 2024-08-19
feat: add e2e test for instructlab CI

RobotSail:
This commit adds a modified version of the E2E NVIDIA T4 workflow from the instructlab/instructlab
repository to automatically test new PRs against an NVIDIA box to see how they perform end-to-end.

This PR follows up work from the following PRs:

- https://github.com/instructlab/training/pull/69
- https://github.com/instructlab/training/pull/77

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E (NVIDIA Tesla T4 x1)
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches:
+      - main
+      - release-*
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+
+jobs:
+  start-runner:
+    name: Start external EC2 runner
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS credentials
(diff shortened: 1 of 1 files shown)


danmcp (approved):
LGTM

RobotSail:
@danmcp Thanks, we already fixed it in another PR so will just need to rebase here.

nathan-weinberg:
@RobotSail i remember you voiced concerns about the E2E script not living in this repo, is that still the case?

RobotSail:
@nathan-weinberg

> i remember you voiced concerns about the E2E script not living in this repo, is that still the case?

Well the current workflow is not ideal but it is less work to have them live there for now until someone is interested in maintaining a separate set of tests in this repo. In the future if we decide to separate repos between upstream & downstream equivalents, the downstream should continue to rely on the tests in the https://github.com/instructlab/instructlab repo. The upstream version, should one exist, would drop this test and would instead rely on its own e2e tests.

nathan-weinberg:
> @nathan-weinberg
> 
> > i remember you voiced concerns about the E2E script not living in this repo, is that still the case?
> 
> Well the current workflow is not ideal but it is less work to have them live there for now until someone is interested in maintaining a separate set of tests in this repo. In the future if we decide to separate repos between upstream & downstream equivalents, the downstream should continue to rely on the tests in the https://github.com/instructlab/instructlab repo. The upstream version, should one exist, would drop this test and would instead rely on its own e2e tests.

SGTM - we could even have a dedicated repo for upstream tests that other repos could pull from - this was done by the OCP org at one point: https://github.com/openshift/openshift-tests

nathan-weinberg (approved):

RobotSail:
> we could even have a dedicated repo for upstream tests that other repos could pull from - this was done by the OCP org at one point: https://github.com/openshift/openshift-tests

Just to be clear, by upstream here you mean upstream relative to the other OCP repos and not upstream as in the upstream Kubernetes, right?

nathan-weinberg:
Yes! So in our case it would be something like:
- a public repo for InstructLab tests that could be used anywhere upstream for testing within library repos (e.g. `instructlab-tests` with different tests that each lib can pull from)
- for tests for downstream projects such as RHEL AI, these tests _could_ be used but the development of them would focus on testing the InstructLab project first and foremost

RobotSail:
Got it, that makes perfect sense.

Issue #148, opened 2024-07-22
Project uses non-inclusive language and problematic language


tiran:
In 2020, Red Hat made a pledge to make open source more inclusive by eradicating problematic language, https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language New projects and new code should make an effort to avoid problematic terms like `master`.

While the env vars `MASTER_ADDR` and `MASTER_PORT` are part of Torch's API, there is really no need to name local variables the same.

RobotSail:
Closing this issue as this is not something we can change - these variables come from the upstream PyTorch project and we wish to be in parity with their naming convention since this simplifies debugging ability and reduces confusion when consuming the training library. All of the variable names that we currently have in the `TorchrunArguments` object are in parity with the argument names to `torchrun`, e.g. `nproc_per_node` is named that when it could instead be `num_gpus` for the sole reason that we don't wish to diverge from upstream.

To address this issue, we need to solve it upstream and then update our downstream code accordingly.

PR #176, opened 2024-08-20
chore: change references to "master" to "main"

nathan-weinberg:
Fixes https://github.com/instructlab/training/issues/148

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,9 +198,9 @@ For example, in a 8-GPU, 2-machine system, we would
 specify the following torchrun config:
 
 ```python
-MASTER_ADDR = os.getenv('MASTER_ADDR')
-MASTER_PORT = os.getnev('MASTER_PORT')
-RDZV_ENDPOINT = f'{MASTER_ADDR}:{MASTER_PORT}'
+MAIN_ADDR = os.getenv('MAIN_ADDR')
+MAIN_PORT = os.getnev('MAIN_PORT')
+RDZV_ENDPOINT = f'{MAIN_ADDR}:{MAIN_PORT}'
 
 # on machine 1
 torchrun_args = TorchrunArgs(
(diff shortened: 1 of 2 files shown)


Maxusmusti (changes requested):
Thanks for the PR @nathan-weinberg! Unfortunately, this will actually break with the current version of pytorch. While we control our local vars, the environment variables are actually set/controlled by torchrun, meaning changing the `os.getenv` calls to look for different env vars will cause them to find nothing. We can update this PR for now just to rename our local variables and leave the env vars as-is, or we can wait for one of the following:
 - upstream torch updates their language
 - we create a fork of torch that changes these fields (likely fragile due to potential requirements to support nvidia pytorch)

nathan-weinberg:
> Thanks for the PR @nathan-weinberg! Unfortunately, this will actually break with the current version of pytorch. While we control our local vars, the environment variables are actually set/controlled by torchrun, meaning changing the `os.getenv` calls to look for different env vars will cause them to find nothing. We can update this PR for now just to rename our local variables and leave the env vars as-is, or we can wait for one of the following:
> 
>     * upstream torch updates their language
> 
>     * we create a fork of torch that changes these fields (likely fragile due to potential requirements to support nvidia pytorch)

I will update the PR to just change the local variable names!

RobotSail:
I don't think that we should update the local variable names, the reason being that it allows us to be consistent with the upstream and makes it easier to debug issues. By staying consistent, we'll make it easier for users and consumers of the training repo to know exactly what each variable is. The `TorchrunArguments` object had its keys specifically designed to be in parity with what `torchrun` uses, e.g. using `nproc_per_node` rather than the easier `num_gpus`.

nathan-weinberg:
@RobotSail That logic is sound - I'm also happy to close this PR but only if we also close the related issue - we need to trim down these backlogs and if we're not going to do the issues they should be closed (like I said I'm fine with either approach!)

RobotSail:
@nathan-weinberg That's totally fine, we can close the other issue then. Thank you for staying on top of the backlog!

nathan-weinberg:
@RobotSail for sure - I will close this - can you close the issue with the explanation you have in the PR so it's tracked there?

RobotSail:
@nathan-weinberg done!

Issue #182, opened 2024-08-26
Issue: Incorrect special tokens for Mistral instruct models


aldopareja:
In `src/instructlab/training/chat_templates/mistral_tmpl.py`, we're currently defining special tokens for user and assistant:

```python
python
SPECIAL_TOKENS = SpecialTokens(
bos="<s>",
eos="</s>",
user="[INST]",
assistant="[/INST]",
)
```

This doesn't align with Mistral's instruct model format, which doesn't use special tokens for user and assistant roles. Instead, Mistral uses `[INST]` and `[/INST]` as part of the prompt template, not as special tokens.

aldopareja:
@Maxusmusti @RobotSail @JamesKunstle

PR #183, opened 2024-08-27
Ap/fix mistral template

aldopareja:
this builds on top of #169, it changes the way we treat system, assistant and user tokens, and assume they are always lists (which could be empty) or single element. We are loosing testing of the alternation between user and assistant. But the jinja template of mistral enforces this, perhaps we could do the same with granite.


aldopareja:
@Maxusmusti @JamesKunstle @RobotSail  this fixes #182

aldopareja:
fixes #184

Maxusmusti:
@aldo-pareja can you rebase this on current main? (which now includes #169 )

Maxusmusti:
Closing as these changes are introduced in #213

PR #188, opened 2024-08-28
add support for CPU and MPS

cdoern:
do not use distributed when not available, instead use CPU or MPS.

This entails a few changes:

* --device is now a valid flag to the library since `ilab` can pass CPU, MPS, or default to cuda when using CPU or MPS, 
* do not initialize DS, instead put the model on the device and initialize `Adafactor` optimizer which is more efficient and than Adam based one
*  inside of `train` add logic for handling if torch.cuda.is_available and torch.distributed.is_initialized() we dont use distributed torch on consumer systems 
* the train loop needs some custom step and loss logic for a LlamaForCausalLM model, add that in when using CPU or MPS we are always world_size == 1 and local_rank == 0

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -16,11 +16,18 @@ from deepspeed.runtime.zero.utils import ZeRORuntimeException
 
 # pylint: disable=no-name-in-module
 from instructlab.dolomite.hf_models import GPTDolomiteForCausalLM
+from torch import nn
 from torch.distributed import ReduceOp, all_reduce
 from tqdm import tqdm
-from transformers import AutoModelForCausalLM, get_scheduler
+from transformers import (
+    Adafactor,
+    AutoModelForCausalLM,
+    LlamaForCausalLM,
+    get_scheduler,
+)
 import deepspeed
 import torch
+import torch.distributed
 
 # First Party
 from instructlab.training import config
@@ -83,7 +90,7 @@ def get_ds_config(world_size, samples_per_gpu, grad_accum, opts: DeepSpeedOption
     return ds_config
 
 
-def setup_model(args, tokenizer, train_loader, grad_accum):
+def setup_model(args, tokenizer, train_loader, grad_accum, device):
     bnb_config = None
     if args.lora_r > 0 and args.lora_quant_bits == 4:
         # Third Party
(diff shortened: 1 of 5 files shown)


RobotSail:
Thank you for this PR, we'll just need to test this to make sure it works with all systems.

cdoern:
e2e is broken because `ilab` doesn't have the changes needed to use this new CPU capable training method.

However, there's no way for me to fix this upstream first bc then ilab CI will break due to not recognizing the training changes since they don't exist yet. 

This seems like a pretty circular dep, I will try to find a way around it

JamesKunstle:
This PR is very valuable. It proves that the existing training loop is minor-modification away from being accelerator-agnostic. However, there are some things that make merging it challenging:

1. The cuda+distributed case are woven together. The vector of acceptable target hardware is: 
[CPU, MPS, Habana, ROCm, Cuda, Habana+Distributed, ROCm+Distributed, Cuda+Distributed].
2. MPS doesn't support quantized training down this route because those layers don't exist in native PyTorch. That means this is a feature regression.
3. Each of the setup components + the training loop become "braided" because we have to handle the two addressed hardware classes separately.

These are criticisms that shouldn't be taken as being heavier-weight than the praise of success for getting this working. Ultimately, I don't think that we **should** merge this now. We're going to be working toward support for two new classes of accelerator, using a new distributed training framework. I think that work will be easier to accomplish if we don't have the outlier CPU and MPS hardware to worry about.

That being said, I don't want to discard this work at all. I'd advocate for it to be "frozen" for a bit until we have more control over the training strategy for the "harder" cases before we adopt these simplest cases.

src/instructlab/training/utils.py
```diff
@@ -28,15 +28,14 @@
...
-from torch import distributed as dist
-from torch.distributed import get_rank, is_initialized
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
     CheckpointImpl,
     apply_activation_checkpointing,
     checkpoint_wrapper,
 )
 import numpy as np
 import torch
+import torch.distributed
```
JamesKunstle, line 38: i'd recommend dropping this refactor. `torch.distributed as dist` is common convention. I would keep `dist.get_rank` and `dist.is_initialized` though, that's clearer.

src/instructlab/training/main_ds.py
```diff
@@ -230,45 +237,56 @@ def make_inputs_require_grad(module, input, output):
...
-    return model
+            lr_scheduler=lr_scheduler,
+            dist_init_required=True,
+        )
+    else:
+        # If we are using CPU or MPS just place model on that device
+        # also, initialize Adafactor, a Transformers Optimizer designed to use less resources.
+        # if we use AdamW here most people will always run out of RAM
+        model = model.to(device)
+        optimizer = Adafactor(
```
JamesKunstle, line 285: we probably shouldn't switch optimizers blindly. AdamW is popular because it's more effective, with its tradeoff. Switching to Adafactor would require some performance experiments.

JamesKunstle:
@cdoern since you've merged CPU + MPS training in `instructlab/instructlab` should this stay open?

JamesKunstle:
@cdoern I'm thinking we should close this since this library, at least in the near term, won't directly support MPS or CPU training. We can revisit this in the future.

PR #189, opened 2024-08-29
ci: minor cleanup of E2E job

nathan-weinberg:
To align with implementation in other library repos

Note `ilab system info` is removed as that is now part of the testing script itself

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -95,6 +95,7 @@ jobs:
       - name: Fetch and checkout PR
         id: fetch_pr
         if: github.event_name == 'pull_request_target'
+        working-directory: "${{ env.GITHUB_WORKSPACE }}/training"
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -64,29 +63,30 @@ jobs:
...
       # for debugging
-      - name: Print Environment State
+      - name: Print environment state
         run: |
           echo "Current Working Directory: $PWD"
           echo "Files in Local Directory:"
           ls -l
-      - name: Checkout
+
+      - name: Checkout instructlab/instructlab
```
JamesKunstle, line 72: do we want to be checking out the current main branch of instructlab if we're testing training? I assume we might prefer to use the most recently published version instead.

nathan-weinberg: Not something that should be changed as part of this PR

JamesKunstle: okay

JamesKunstle (approved):
tests still pass and it's an in-place rename

danmcp (approved):

PR #192, opened 2024-08-30
Fix linting error and missing quote

danmcp:
Ex of error: https://github.com/instructlab/training/actions/runs/10624704122/job/29453541657?pr=192


RobotSail (approved):
LGTM

PR #212, opened 2024-09-18
Ap/accelerate fsdp

aldopareja:
accelerate has been tested to run with parity check between the fsdp and deepspeed backends. Also created a function to do the saving. Need to test:

- QLora
- LoRA
- Checkpointing (although accelerate has an API for that)


Maxusmusti:
Closing as #213 introduces these changes

PR #215, opened 2024-09-18
test failure 2

RobotSail:
this tests if the e2e correctly fails within the training loop


JamesKunstle:
@RobotSail can we close this?

RobotSail:
@JamesKunstle Yes

PR #229, opened 2024-09-26
changes StreamablePopen to return a process and implement listening

JamesKunstle:
The original logic wasn't working because StreamablePopen started a process and then listened on the output file instead of returning a process that could be terminated/killed.

This adds 'process.listen' to implement the same behavior as before, but only after the StreamablePopen object has returned a reference to itself. When a KeyboardInterrupt is sent, the process can be terminated and/or killed, and the Exception/Interrupt can be raised.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -726,32 +726,45 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         f"--fsdp_sharding_strategy={train_args.fsdp_options.sharding_strategy.value}"
     )
 
-    print(f"\033[92mRunning command: {' '.join(command)}\033[0m")
+    print(f"\033[92mRunning training command as subprocess: {' '.join(command)}\033[0m")
     process = None
+    interrupt: KeyboardInterrupt | Exception | None = None
     try:
         process = StreamablePopen(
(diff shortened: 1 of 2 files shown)


RobotSail (approved):
LGTM

src/instructlab/training/main_ds.py
```diff
@@ -4,6 +4,7 @@
 from copy import deepcopy
 from pathlib import Path
 import argparse
+import logging
```
RobotSail, line 7: Have you tested that this logging works? If I remember correctly, you need to initialize the logger first before it prints anything. Otherwise calling `logging.xxx` doesn't do anything.

Maxusmusti (changes requested):
Fix @RobotSail suggestion

src/instructlab/training/main_ds.py
JamesKunstle: setup logger

RobotSail: Nice, thank you!

JamesKunstle:
@Maxusmusti @RobotSail removed `logging`, back to printing.

RobotSail:
Thanks! LGTM

Maxusmusti (approved):
LGTM

PR #235, opened 2024-10-01
adds Accelerate full-state (opt, lr_sched, params)

JamesKunstle:
Saving and reloading for deepspeed.

Todo:

- [ ] Need to save LoRA model correctly
- [ ] choose whether the save the transformers model config + tokenizer with the model.
- [ ] make sure this also handles dolomite correctly
- [ ] run w/ CPU offloading
- [ ] choose how to resume w/ samples_seen and between epochs. e.g. if we save a checkpoint at the end of an epoch, do we always save with the metadata for the next epoch and samples_seen=0?

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -43,9 +43,11 @@ from instructlab.training.utils import (
     apply_gradient_checkpointing,
     convert_loss_to_reduce_sum,
     ensure_loadable_granite_checkpoint,
+    load_latest_full_state,
     prepare_peft_model,
     prepare_universal_checkpoint_from_latest,
     retrieve_chat_template,
+    save_checkpoint,
     save_hf_format_accelerate,
     set_random_seed,
     setup_logger,
@@ -316,6 +318,10 @@ def train(
     batch_size = args.effective_batch_size // grad_accum
     samples_seen = 0
 
+    if hasattr(args, "samples_seen"):
+        print(f"\033[93mUpdating 'samples_seen' {args.samples_seen}\033[0m")
+        samples_seen = args.samples_seen
+
     if args.save_samples > 0:
         args.save_samples = (args.save_samples // batch_size) * batch_size
         (
@@ -335,7 +341,7 @@ def train(
         )
 
     global_grad_norm = None
-    for epoch in range(args.num_epochs):
+    for epoch in range(args.current_epoch, args.num_epochs):
         if args.sampler in ("multipack"):
             train_loader.batch_sampler.set_epoch(epoch)
         elif args.sampler in ("distributed"):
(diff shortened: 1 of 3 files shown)


src/instructlab/training/main_ds.py
```diff
@@ -335,7 +337,7 @@ def train(
         )
 
     global_grad_norm = None
-    for epoch in range(args.num_epochs):
+    for epoch in range(args.current_epoch, args.num_epochs):
```
RobotSail, line 340: Needs to be 
```suggestion
    for epoch in range(args.current_epoch, args.num_epochs+1):
```
 
Since `range(start, stop)` only goes from `start, ..., stop - 1`

src/instructlab/training/utils.py
```diff
@@ -781,3 +782,103 @@ def set_random_seed(seed):
...
+    in `output_dir/full_state`. 
+    """
+    output_dir = Path(args.output_dir) / "full_state"
+
+    # dist.breakpoint()
+    if not output_dir.is_dir():
+        return
+
+    # picks checkpoint with the largest number of samples seen, by name.
+    checkpoint_list = sorted(list(output_dir.iterdir()), reverse=True)
```
RobotSail, line 864: Is this necessarily true? If the list is a list of strings then how does it handle a case like `samples_1000` vs `samples_99999`

src/instructlab/training/utils.py
```diff
@@ -781,3 +782,103 @@ def set_random_seed(seed):
...
+
+    if len(checkpoint_list) == 0:
+        log_rank_0(f"\033[93mNo checkpoints to load from: {output_dir}\033[0m", to_print=True)
+        return
+
+
+    latest = checkpoint_list[0]
```
RobotSail, line 871: Are we guaranteed to have at least one element when we check this?

RobotSail (commented):
I left a few comments but overall it looks good

src/instructlab/training/utils.py
JamesKunstle: Usually hash of the string or sum of the string characters, so 9999 > 1000.

src/instructlab/training/utils.py
JamesKunstle: Yeah because the check above checks for len 0

src/instructlab/training/utils.py
RobotSail: I would advise passing a custom comparison function to sorted here, see the following code snippet:

```py
lst = ["hf_format_1", "hf_format_10", "hf_format_2"]
sorted_lst = sorted(lst)
print(sorted_lst)
> ['hf_format_1', 'hf_format_10', 'hf_format_2']
```

You can pass a custom function that simply splits on the `_` character and casts the last part of the string into a number, this should do what you want it to.

RobotSail: This will get `sorted` to work how you want it to:
```suggestion
    checkpoint_list = sorted(list(output_dir.iterdir()), reverse=True, key=lambda d: int(d.split('_')[-1]))
```

Maxusmusti (approved):
EL GEE TEE EM

PR #257, opened 2024-10-07
Implementing HF Padding-Free and GraniteLM Support

aldopareja:
Updating the data collator for models with HF padding-free support, adding support for upcoming Granite HF model class, and updating flags/interface accordingly.

 -Mustafa

### DIFF
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -354,10 +354,12 @@ def main(args: DataProcessArgs):
         num_proc=NUM_PROC,
     )
     data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
+    # MASK and both pretrain tokens should not be in the final tokens, those are special tokens added only for data processing purposes.
     max_id = len(tokenizer) - 3
     final_valid_data = data_with_labels.filter(
         lambda x: all(tk < max_id for tk in x["labels"]), num_proc=NUM_PROC
     )
+    # Dropping samples that could break training due to oob ids
     if len(final_valid_data) < len(data_with_labels):
         dropped_samples = len(data_with_labels) - len(final_valid_data)
         print(
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -348,7 +348,10 @@ def main(args: DataProcessArgs):
 
     # extract only labels and messages formatted into a new dataset
     data_with_labels = data_with_labels.map(
-        lambda x: {"len": len(x["input_ids"]), }, num_proc=NUM_PROC
+        lambda x: {
+            "len": len(x["input_ids"]),
+        },
+        num_proc=NUM_PROC,
     )
     data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
     # use path to get the stem of the file
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -347,7 +347,10 @@ def main(args: DataProcessArgs):
     )
 
     # extract only labels and messages formatted into a new dataset
-    data_with_labels = data_with_labels.select_columns(["labels", "input_ids"])
+    data_with_labels = data_with_labels.map(
+        lambda x: {"len": len(x["input_ids"]), }, num_proc=NUM_PROC
+    )
+    data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
     # use path to get the stem of the file
     data_with_labels.to_json(Path(args.data_output_path) / f"data.jsonl")
 
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -354,6 +354,15 @@ def main(args: DataProcessArgs):
         num_proc=NUM_PROC,
     )
     data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
+    max_id = len(tokenizer) - 3
+    final_valid_data = data_with_labels.filter(
+        lambda x: all([tk < max_id for tk in x["labels"]]), num_proc=NUM_PROC
+    )
+    if len(final_valid_data) < len(data_with_labels):
+        dropped_samples = len(data_with_labels) - len(final_valid_data)
+        print(
+            f"\033[93mWarning: {dropped_samples} samples were dropped because they contained token IDs greater than or equal to {max_id}.\033[0m"
+        )
     # use path to get the stem of the file
     data_with_labels.to_json(Path(args.data_output_path) / "data.jsonl")
 
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -355,7 +355,7 @@ def main(args: DataProcessArgs):
     )
     data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
     # use path to get the stem of the file
-    data_with_labels.to_json(Path(args.data_output_path) / f"data.jsonl")
+    data_with_labels.to_json(Path(args.data_output_path) / "data.jsonl")
 
 
 if __name__ == "__main__":
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -356,7 +356,7 @@ def main(args: DataProcessArgs):
     data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
     max_id = len(tokenizer) - 3
     final_valid_data = data_with_labels.filter(
-        lambda x: all(tk < max_id for tk in x['labels']), num_proc=NUM_PROC
+        lambda x: all(tk < max_id for tk in x["labels"]), num_proc=NUM_PROC
     )
     if len(final_valid_data) < len(data_with_labels):
         dropped_samples = len(data_with_labels) - len(final_valid_data)
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -199,7 +199,7 @@ def print_masked_samples(data, tokenizer, is_pretrain, num_proc):
     def get_masked_and_orig_text(sample):
         labels = sample["labels"]
         input_ids = sample["input_ids"]
-        mask_id = get_sp_token(tokenizer, "<MASK>")[0]
+        mask_id = get_sp_token(tokenizer, "<|MASK|>")[0]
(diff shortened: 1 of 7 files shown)


src/instructlab/training/token_dataset.py
```diff
@@ -11,18 +11,25 @@
...
-        )
+        if "len" not in self.data.column_names:
+            self.lengths = np.array(
+                self.data.map(
+                    lambda x: {"len": len(x["input_ids"])},
+                    num_proc=8,
+                )["len"]
+            )
+        else:
+            self.lengths = np.array(self.data["len"])
```
RobotSail, line 32: I'm a bit unfamiliar with setting up datasets like this - is there a reason we'd want to set this up as a numpy array first instead of just using `torch.tensor` directly?

src/instructlab/training/utils.py
```diff
@@ -111,9 +111,18 @@ def listen(self):
...
 
 
-def make_collate_fn(pad_token_id, is_granite=False, max_batch_len=60000):
+def supports_flash_attention(device_id=0):
+    """Check if a GPU supports FlashAttention."""
+    major, minor = torch.cuda.get_device_capability(device_id)
+    # Check if the GPU architecture is Ampere (SM 8.x) or newer (SM 9.0)
+    is_sm8x = major == 8 and minor >= 0
+    is_sm90 = major == 9 and minor == 0
+    return is_sm8x or is_sm90
```
RobotSail, line 120: What happens if someone runs this on AMD, would this be able to handle that case as well?

src/instructlab/training/utils.py
```diff
@@ -140,70 +149,108 @@ def pad_collate_fn(batch):
...
+                input_ids = []
+                labels = []
+                position_ids = []
+                total_len = 0
+                num_loss_counted_tokens = 0
+
+                for num_samples, item in enumerate(batch):
+                    item_len = len(item["input_ids"])
+                    if total_len + item_len > max_batch_len:
+                        break
```
RobotSail, line 168: Could this potentially drop samples if we're breaking after taking a sample from the batch without putting it back?

src/instructlab/training/utils.py
Maxusmusti: No, I'll need to update this to handle it

src/instructlab/training/token_dataset.py
aldopareja: to get the mean() with a single command later on. Prefer numpy to make sure it's in CPU.

src/instructlab/training/utils.py
aldopareja: yes but it doesn't matter, and also, with multipack it never goes beyond max batch len so this never happens.

src/instructlab/training/data_process.py
```diff
@@ -347,9 +347,24 @@ def main(args: DataProcessArgs):
...
     # extract only labels and messages formatted into a new dataset
-    data_with_labels = data_with_labels.select_columns(["labels", "input_ids"])
+    data_with_labels = data_with_labels.map(
+        lambda x: {
+            "len": len(x["input_ids"]),
+        },
+        num_proc=NUM_PROC,
+    )
+    data_with_labels = data_with_labels.select_columns(["labels", "input_ids", "len"])
+    max_id = len(tokenizer) - 3
```
JamesKunstle, line 357: what is the -3 for?

src/instructlab/training/data_process.py
```diff
@@ -347,9 +347,24 @@ def main(args: DataProcessArgs):
...
+    final_valid_data = data_with_labels.filter(
+        lambda x: all(tk < max_id for tk in x["labels"]), num_proc=NUM_PROC
+    )
+    if len(final_valid_data) < len(data_with_labels):
+        dropped_samples = len(data_with_labels) - len(final_valid_data)
+        print(
+            f"\033[93mWarning: {dropped_samples} samples were dropped because they contained token IDs greater than or equal to {max_id}.\033[0m"
```
JamesKunstle, line 364: can this be a problem ever? I imagine that any unaccounted-for tokens would be assigned to UNK but maybe the dropped tokens (from the above filtering step) should be logged just in case.

src/instructlab/training/main_ds.py
```diff
@@ -102,15 +104,11 @@ def setup_model(args, tokenizer, train_loader, grad_accum):
         "torch_dtype": torch.bfloat16,
         "quantization_config": bnb_config,
     }
-    if not args.disable_flash_attn:
+    if flash_enabled:
         base_model_args["attn_implementation"] = "flash_attention_2"
```
JamesKunstle, line 108: is `flash_attention_2` going to be coming from the `FlashAttention` repo or is this something that will need to be built by core components?

src/instructlab/training/main_ds.py
```diff
@@ -639,28 +645,26 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
...
-                #
-                # An important reason for why #1 would be preferable is in the case of OpenShift/SELinux
-                # where the user has a defined place for new temporary data to be written.
-                data_output_path=train_args.data_output_dir,
-                model_path=train_args.model_path,
-                data_path=train_args.data_path,
-                max_seq_len=train_args.max_seq_len,
-                chat_tmpl_path=train_args.chat_tmpl_path,
-            )
+    check_valid_train_args(train_args)
```
JamesKunstle, line 648: I think this removes the optional data processing check on line 648 that @MichaelClifford added a while ago. This should be kept.

src/instructlab/training/utils.py
```diff
@@ -40,6 +41,44 @@
...
+    # early validation logic here
+    if train_args.max_batch_len < train_args.max_seq_len:
+        raise ValueError(
+            f"the `max_batch_len` cannot be less than `max_seq_len`: {train_args.max_batch_len=} < {train_args.max_seq_len=}"
+        )
+
+    if os.path.exists(train_args.model_path):
+        if not os.path.isdir(train_args.model_path):
+            raise FileNotFoundError(
+                "Model path does not appear to be a dir, please validate or update the path"
```
JamesKunstle, line 58:
```suggestion
                "Model path does not appear to be a directory. Please make sure that you're passing a Hugging Face Transformers compatible directory checkpoint."
```

src/instructlab/training/utils.py
```diff
@@ -40,6 +41,44 @@
...
+            )
+    else:
+        raise FileNotFoundError(
+            "Model Path cannot be found, please verify existence and permissions"
```
JamesKunstle, line 62:
```suggestion
            f"Provided path to model does not exist. Please make sure that you've passed a valid model and that it has appropriate permissions: {train_args.model_path}"
```

JamesKunstle (commented):
Generally seems like a solid drop-in replacement with some good updates. First round of review only covers code- I'll run everything on AMD hardware as well.

src/instructlab/training/main_ds.py
Maxusmusti: Oh good catch, let me fix this

src/instructlab/training/main_ds.py
Maxusmusti: This is a model arg available during model instantiation (via transformers), if included it will import and utilize the flash attention package

src/instructlab/training/main_ds.py
Maxusmusti: ok fixed

src/instructlab/training/utils.py
Maxusmusti: updated

src/instructlab/training/utils.py
Maxusmusti: updated

src/instructlab/training/data_process.py
aldopareja: because MASK and both pretrain tokens should not be in the tokens, those are special tokens added only for data processing purposes.

src/instructlab/training/data_process.py
aldopareja: yes, a dataset I was training on had the same string that we used for the MASK token, then the integer that would be mapping to would be higher than the integers the model would have and the training would break. This is only for the tokens that are added here but not in training.

src/instructlab/training/main_ds.py
JamesKunstle: 👍

src/instructlab/training/data_process.py
JamesKunstle: Could you please add an explanation comment for that? Raw integers like that are called "magic numbers" and are usually managed by either naming a constant to replace it or commenting what it means.

Maxusmusti: I'll do that

src/instructlab/training/data_process.py
Maxusmusti: I'll do that

Maxusmusti: done

src/instructlab/training/data_process.py
Maxusmusti: done

Maxusmusti:
@JamesKunstle added comments for the things you wanted

JamesKunstle:
@Maxusmusti Much appreciated- testing on AMD now

JamesKunstle:
Testing passes on AMD, losses between Dolomite and Llama are in parity.

JamesKunstle (approved):
Tests pass and loss curves look good between Dolomite and Llama padding-free on AMD. Feel good about merging this.

src/instructlab/training/data_process.py
```diff
@@ -199,7 +199,7 @@ def print_masked_samples(data, tokenizer, is_pretrain, num_proc):
     def get_masked_and_orig_text(sample):
         labels = sample["labels"]
         input_ids = sample["input_ids"]
-        mask_id = get_sp_token(tokenizer, "<MASK>")[0]
+        mask_id = get_sp_token(tokenizer, "<|MASK|>")[0]
```
RobotSail, line 202: Will this affect existing models? Or is this purely for training-time?

RobotSail (approved):
LGTM

src/instructlab/training/data_process.py
Maxusmusti: Yeah only relevant during training

Issue #254, opened 2024-10-07
Division by zero error when calling `find_packing_max_batch_len_and_grad_accum`


RobotSail:
It seems like there's a division by zero error that can still occur with the right conditions.

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/oleg/Programming/training/src/instructlab/training/main_ds.py", line 934, in <module>
[rank0]:     main(args)
[rank0]:   File "/home/oleg/Programming/training/src/instructlab/training/main_ds.py", line 546, in main
[rank0]:     packing_max_batch_len, grad_accum = find_packing_max_batch_len_and_grad_accum(
[rank0]:                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/oleg/Programming/training/src/instructlab/training/multipack_sampler.py", line 160, in find_packing_max_batch_len_and_grad_accum
[rank0]:     packing_max_batch_len = find_max_pack_len_with_padding(
[rank0]:                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/oleg/Programming/training/src/instructlab/training/multipack_sampler.py", line 86, in find_max_pack_len_with_padding
[rank0]:     avg_bs_per_minibatch = get_effective_samples_per_minibatch(
[rank0]:                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/oleg/Programming/training/src/instructlab/training/multipack_sampler.py", line 79, in get_effective_samples_per_minibatch
[rank0]:     return len(dataset) / len(batches)
[rank0]:            ~~~~~~~~~~~~~^~~~~~~~~~~~~~
[rank0]: ZeroDivisionError: division by zero
```

To reproduce: 

1. Run a model train with the [community dataset](https://huggingface.co/datasets/instructlab/InstructLabCommunity)
2. Set your `TrainingArgs` like so:

```
train_args = TrainingArgs(
    data_path="path/to/data",
    is_padding_free=False,
    ckpt_output_dir='checkpoints',
    data_output_dir='/dev/shm',
    max_seq_len=70,
    max_batch_len=70,
    effective_batch_size=3840,
    save_samples=20_000,
)
```

JamesKunstle:
@RobotSail I think you fixed this!

PR #273, opened 2024-10-15
Fix for https://github.com/instructlab/training/issues/254

ashna000:
Added checks for resolving the issue reported in https://github.com/instructlab/training/issues/254


ashna000:
@RobotSail Can you please review this ?

src/instructlab/training/multipack_sampler.py
```diff
@@ -76,7 +76,7 @@ def get_effective_samples_per_minibatch(num_tokens_per_gpu):
             padding=True,
         )
         batches = sampler.generate_batches()
-        return len(dataset) / len(batches)
+        return len(dataset) / len(batches) if len(batches) > 0 else None
```
RobotSail, line 79: If we're hitting a scenario where Multipack generated no batches, then we need to either error out and have the calling function handle it (by falling back to `DistributedDataSampler`, or simply by preventing us from using multipack in the first place.

My recommendation here is to do this:
- Raise an appropriate exception here (either `RuntimeError` or `DivisionByZeroError`) 
- Have the calling function check for this additional exception and fall back to the naive sampler

RobotSail:
Thanks for this PR @ashna000, this will be a really important fix ! I've left a comment on how we should handle this problem.

nathan-weinberg (changes requested):
@ashna000 can you please squash your commits?

nathan-weinberg:
@ashna000 these still need to be squashed

PR #280, opened 2024-10-17
adds basic smoketests for main_ds and data_process CLI args

JamesKunstle:
During development it's convenient to be able to run full distributed training, even on a smaller dataset, just to make sure that nothing obviously fails. This will also capture support for flash attention on the machine that it's run on, and for granite models.

Also solves an inconvenience that's been annoying- testing for a given platform involves:
1. grab image with versions of dependencies for cards
2. run container
3. download and install this repo
4. write dumb versions of these scripts to run training and see if it breaks.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,20 @@
+## Overview
+
+`smoketest.sh` cd's into the source directory and runs the script-entrypoint for `main_ds.py` and `data_process.py`. Testing will break if file names or locations in the source tree change.
+
+Existing tests are "smoke tests," meant to demonstrate that training completes (returns 0) or not. This is helpful to check if all required dependencies are installed.
+
+Current tests add features as they go:
+
+1. No Flash Attention or Granite
+2. No Granite but Flash Attention enabled
+3. Granite and Flash Attention enabled
+
+## Usage
+
+The testing script can be run without parameters as `./smoketest.sh`. By default, this will run all tests with `FSDP` as the distributed training backend. To change the distributed training backend to the other available option, one can run the script as `./smoketest.sh deepspeed`.
+
+The second positional argument is for "number of GPUs"- e.g.: `./smoketest.sh fsdp 8`. This will run the test with 8 GPUs with fsdp as the distributed backend.
+
+> [!NOTE]
+> You'll need to install the training library to run the test. Inside a virtual environment and at inside the repo, please run `pip3 install -e .` to install the package in editable mode.
--- a/tests/smoketest.sh
+++ b/tests/smoketest.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+
+# ############### Read-only parameters ############### 
+MODEL_NAME="instructlab/granite-7b-lab"
+# gets directory of current file.
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+CORRECT_WORKING_DIR="${SCRIPT_DIR}/../src/instructlab/training/"
+SAMPLE_DATA_PATH="${SCRIPT_DIR}/../sample-data/train_all_pruned_SDG.jsonl"
+TMP_DIR=$(mktemp -d)
+CHECKPOINTS_DIR="${TMP_DIR}/checkpoints"
+DATA_DIR="${TMP_DIR}/data"
+COMPUTED_DATA_PATH="${DATA_DIR}/data.jsonl"
+DEFAULT_DISTRIB_FRAMEWORK='fsdp'
+DISTRIB_FRAMEWORK="${1:-$DEFAULT_DISTRIB_FRAMEWORK}" # defaults to FSDP
+DEFAULT_GPUS=8
+NUM_GPUS="${2:-$DEFAULT_GPUS}"
+
+# ############### User-modifiable parameters ############### 
+# Change these as needed
+MAX_BATCH_LEN=60000
+NUM_SAMPLES_TRAINED_ON=5000 # upper-bound on training dataset size.
+
+# ############### Test Functions ############### 
+
+#######################################
+# Creates directories for the precomputed datasets
+# and the checkpoints that are saved during training inside
+# of the temporary storage created for these tests.
+# Globals:
+#   CHECKPOINTS_DIR 
+#   DATA_DIR
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+function setup_tmpdir () {
+    mkdir "$CHECKPOINTS_DIR"
+    mkdir "$DATA_DIR"
+}
+
+#######################################
+# Test most common training parameters without using
+# Flash Attention
+# Globals:
+#   SAMPLE_DATA_PATH
+#   DATA_DIR
+#   MODEL_NAME
+#   NUM_SAMPLES_TRAINED_ON
+#   COMPUTED_DATA_PATH
+# Arguments:
+#   None
+# Returns:
+#   echos number of samples trained on to standard out.
+#######################################
+function prepare_data () {
+    # preprocesses .jsonl messages data so that it's a valid
+    # input to the model (inputs tokenized, formatted with mask, etc.)
+    # then, data is trimmed to a determined length to make training
+    # go faster.
+    
+    python3 data_process.py \
+    --data_path="$SAMPLE_DATA_PATH" \
+    --data_output_path="$DATA_DIR" \
+    --max_seq_len=4096 \
+    --model_name_or_path="$MODEL_NAME"
+
+    # trim data so we only keep the first 'n' samples.
+    # should be enough data for training to be meaningful but not enough
+    # that training takes a large amount of time.
+    echo "$(head -"$NUM_SAMPLES_TRAINED_ON" "$COMPUTED_DATA_PATH")" > "$COMPUTED_DATA_PATH"
+
+    echo "TRAINING ON $(wc -l "$COMPUTED_DATA_PATH") SAMPLES"
+}
+
+#######################################
+# Clears and remakes the temporary directory where
+# artifacts, such as checkpoints and logs, are stored
+# during training.
+# Globals:
+#   CHECKPOINTS_DIR
+# Arguments:
+#   None
+# Returns:
+#   writes location of checkpoints dir to standard out.
+#######################################
+function _cleanup_saved_checkpoints() {
+    echo "CLEARING CHECKPOINTS: $CHECKPOINTS_DIR"
+    rm -rf "$CHECKPOINTS_DIR"
+    mkdir "$CHECKPOINTS_DIR"
+}
+
+#######################################
+# Test most common training parameters without using
+# Flash Attention
+# Globals:
+#   NUM_GPUS
+#   MODEL_NAME
+#   COMPUTED_DATA_PATH
+#   CHECKPOINTS_DIR
+#   DISTRIBUTED_FRAMEWORK
+#   MAX_BATCH_LEN
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+function test_standard_loop () {
+    torchrun \
+    --standalone \
+    --nproc_per_node="$NUM_GPUS" \
+    main_ds.py \
+    --model_name_or_path="$MODEL_NAME" \
+    --data_path="$COMPUTED_DATA_PATH" \
+    --output_dir="$CHECKPOINTS_DIR" \
+    --num_epochs=1 \
+    --effective_batch_size=128 \
+    --save_samples=0 \
+    --checkpoint_at_epoch \
+    --accelerate_full_state_at_epoch \
+    --distributed_training_framework="$DISTRIB_FRAMEWORK" \
+    --max_batch_len="$MAX_BATCH_LEN" \
+    --is_granite
+}
+
+#######################################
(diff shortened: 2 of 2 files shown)


tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
...
+# Change these as needed
+NUM_GPUS=8
+MAX_BATCH_LEN=60000
+NUM_SAMPLES_TRAINED_ON=5000 # upper-bound on training dataset size.
+
+# ############### Read-only parameters ############### 
+MODEL_NAME="instructlab/granite-7b-lab"
+# gets directory of current file.
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+CORRECT_WORKING_DIR="$SCRIPT_DIR/../src/instructlab/training/"
```
RobotSail, line 14: @JamesKunstle Could you please use `"${SCRIPT_DIR}"` as the format for referencing shell variables? This will be really useful to have as a consistent format and reduces potential bugs introduced through incorrect delimiting.

tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
...
+    --model_name_or_path="$MODEL_NAME" \
+    --data_path="$COMPUTED_DATA_PATH" \
+    --output_dir="$CHECKPOINTS_DIR" \
+    --num_epochs=1 \
+    --effective_batch_size=128 \
+    --save_samples=0 \
+    --checkpoint_at_epoch \
+    --accelerate_full_state_at_epoch \
+    --distributed_training_framework="$DISTRIB_FRAMEWORK" \
+    --max_batch_len="$MAX_BATCH_LEN" \
```
RobotSail, line 102: This will cause it to break if we keep the training `\`
```suggestion
    --max_batch_len="$MAX_BATCH_LEN"
```

tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
...
+
+# ############### Test Functions ############### 
+
+function setup_tmpdir () {
+    mkdir "$CHECKPOINTS_DIR"
+    mkdir "$DATA_DIR"
+}
+
+function prepare_data () {
+    # preprocesses .jsonl messages data so that it's a valid
```
RobotSail, line 31: For documenting shell functions, the [recommendation for documenting functions](https://google.github.io/styleguide/shellguide.html#s4.2-function-comments) would be:

```bash
################################################################
# Preprocesses .jsonl messages data so that it's a valid
# input to the model (inputs tokenized, formatted with mask,
# etc.)
# then, data is trimmed to a determined length to make training
# go faster.
# Inputs:
#     None
# Globals:
#     - SAMPLE_DATA_PATH
#     - DATA_DIR
#     - MODEL_NAME
#     - SAMPLES_TRAINED_ON
#     - COMPUTED_DATA_PATH
# Returns:
#     None
################################################################

```

RobotSail: ^ We should adopt the above convention so that it's easier for people to drop in and build on top of these tests in the future. Otherwise they will need to spend a lot more effort to understand why all of these things exist and what their purpose is.

tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
...
+    --checkpoint_at_epoch \
+    --accelerate_full_state_at_epoch \
+    --distributed_training_framework="$DISTRIB_FRAMEWORK" \
+    --max_batch_len="$MAX_BATCH_LEN" \
+    --disable_flash_attn
+    # --is_granite \
+}
+
+# ############### Setup and tests ############### 
+setup_tmpdir
```
RobotSail, line 132: We should probably move this into a `main` function:

```suggestion
function main() {
    setup_tmpdir
    trap "rm -rf $TMP_DIR" EXIT
    
    #NOTE (jkunstle): script is run as though it's
    # in the same source dir as main_ds and data_process.
    cd "$CORRECT_WORKING_DIR"
    echo "CURRENT WORKING DIRECTORY: $(pwd)"
    
    prepare_data
    test_standard_loop_noflashattention_nogranite
    _cleanup_saved_checkpoints
    test_standard_loop_nongranite
    _cleanup_saved_checkpoints
    test_standard_loop
}

main
```

tests/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+`smoketest.sh` cd's into the source directory and runs the script-entrypoint for `main_ds.py` and `data_process.py`. Testing will break if file names or locations in the source tree change.
+
+Existing tests are "smoke tests," meant to demonstrate that training completes (returns 0) or not. This is helpful to check if all required dependencies are installed.
+
+Tests add features over time:
+
+1. No Flash Attention or Granite
+2. No Granite but Flash Attention enabled
+3. Granite and Flash Attention enabled
```
RobotSail, line 11: We should consider making this a checkmark list so we can check things off over time:

```md
- [ ] No Flash Attention or Granite
- [ ] No Granite but Flash Attention enabled
- [ ] Granite and Flash Attention enabled
```

tests/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Usage
+
+The testing script can be run without parameters as `./smoketest.sh`. By default, this will run all tests with `FSDP` as the distributed training backend. To change the distributed training backend to the other available option, one can run the script as `./smoketest.sh deepspeed`.
+
+> NOTE: You'll need to install the training library to run the test. Inside a virtual environment and at inside the repo, please run `pip3 install -e .` to install the package in editable mode.
```
RobotSail, line 17: Github actually has a syntax for this built-in:
```suggestion
> [!NOTE]
> You'll need to install the training library to run the test. Inside a virtual environment and at inside the repo, please run `pip3 install -e .` to install the package in editable mode.
```

It renders like this:

> [!NOTE]
> You'll need to install the training library to run the test. Inside a virtual environment and at inside the repo, please run `pip3 install -e .` to install the package in editable mode.

tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env bash
+set -eux
```
RobotSail, line 2: I would also add `-o pipefail` here as this will cause it to stop if any command within a pipeline fails:
```suggestion
set -eux -o pipefail
```

RobotSail (commented):
Looks good, some minor comments. Once they are addressed I will LGTM

RobotSail:
The main thing is that to make it easier for new developers to come on board and build on top of what we, we should standardize on the [Google Shell Styleguide](https://google.github.io/styleguide/shellguide.html). This will make it much easier for newcomers to understand how things work, why they're organized how they are, and make changes with confidence.

tests/smoketest.sh
JamesKunstle: 👍

tests/smoketest.sh
JamesKunstle: 👍

tests/smoketest.sh
JamesKunstle: 👍

tests/README.md
JamesKunstle: These are done

tests/README.md
JamesKunstle: 👍

tests/smoketest.sh
JamesKunstle: 👍

tests/smoketest.sh
```diff
@@ -0,0 +1,145 @@
...
+
+# ############### User-modifiable parameters ############### 
+# Change these as needed
+NUM_GPUS=8
```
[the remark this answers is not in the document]
JamesKunstle: 👍

tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+`smoketest.sh` cd's into the source directory and runs the script-entrypoint for `main_ds.py` and `data_process.py`. Testing will break if file names or locations in the source tree change.
+
+Existing tests are "smoke tests," meant to demonstrate that training completes (returns 0) or not. This is helpful to check if all required dependencies are installed.
+
+Current tests add features as they go:
+
+1. No Flash Attention or Granite
+2. No Granite but Flash Attention enabled
+3. Granite and Flash Attention enabled
```
Maxusmusti, line 11: Do we want a fourth Dolomite path? I guess this might make more sense after #257

tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Usage
+
+The testing script can be run without parameters as `./smoketest.sh`. By default, this will run all tests with `FSDP` as the distributed training backend. To change the distributed training backend to the other available option, one can run the script as `./smoketest.sh deepspeed`.
+
+The second positional argument is for "number of GPUs"- e.g.: `./smoketest.sh fsdp 8`. This will run the test with 8 GPUs with fsdp as the distributed backend.
+
```
Maxusmusti, line 18: This is a little confusing. We have FSDP as the default, so I can use `smoketest.sh` as-is. But if I want to set the number of GPUs, do I also have to specify FSDP? Since it is the "second positional argument"? Will it break if I just do `smoketest.sh 8`? We may want to instead just have like `--backend/-b and --num-gpus/-n`

tests/smoketest.sh
```diff
@@ -0,0 +1,212 @@
...
+
+    prepare_data
+    test_standard_loop_noflashattention_nogranite
+    _cleanup_saved_checkpoints
+    test_standard_loop_nongranite
+    _cleanup_saved_checkpoints
+    test_standard_loop
+}
+
+main
```
Maxusmusti, line 212: missing newline at EOF

tests/README.md
JamesKunstle: Yeah that'd be smart, we can add that in another iteration

tests/README.md
JamesKunstle: Yeah it's a little annoying because a bit more boilerplate is required to have that behavior. I would prefer for this to be an issue + feature rather than adding it to this PR just because I want this test to be available asap for some testing stuff elsewhere. Would that be chill or do you want the behavior now?

tests/smoketest.sh
JamesKunstle: ++ will fix

tests/README.md
JamesKunstle: created an issue for this

tests/README.md
JamesKunstle: created an issue for this

tests/README.md
RobotSail: Even better

tests/smoketest.sh
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+
+# ############### Read-only parameters ############### 
+MODEL_NAME="instructlab/granite-7b-lab"
```
RobotSail, line 5: Just FYI - the BASH convention is to use `'` for constant-value strings and `"` for strings which you interpolate variables into. Not a big deal here, just something to be aware of.

tests/smoketest.sh
```diff
@@ -0,0 +1,212 @@
...
+# Globals:
+#   CHECKPOINTS_DIR 
+#   DATA_DIR
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+function setup_tmpdir () {
+    mkdir "$CHECKPOINTS_DIR"
```
RobotSail, line 39: It may be wise here to use `mkdir -p` in case the parent directories don't exist for whatever reason.

tests/smoketest.sh
RobotSail: Thank you for doing this, it looks great now!

tests/smoketest.sh
```diff
@@ -0,0 +1,212 @@
...
+    torchrun \
+    --standalone \
+    --nproc_per_node="$NUM_GPUS" \
+    main_ds.py \
+    --model_name_or_path="$MODEL_NAME" \
+    --data_path="$COMPUTED_DATA_PATH" \
+    --output_dir="$CHECKPOINTS_DIR" \
+    --num_epochs=1 \
+    --effective_batch_size=128 \
+    --save_samples=0 \
```
RobotSail, line 119: Would it make sense to have this set to a non-zero value so the relevant code-path  is tested? I can imagine our spaghetti saving logic potentially breaking and this not picking up on it.

RobotSail (approved):
LGTM, a few comments that should be looked at but nothing blocking. Great work!

tests/smoketest.sh
JamesKunstle: ++

tests/smoketest.sh
JamesKunstle: That particular dir will always have extant parents because it's directly descended from the temp dir, so this would appropriately break if something doesn't work

tests/smoketest.sh
JamesKunstle: That's a good point, we can consider that in another test- this is data-dependent so we might want to tie it to the number of samples.

tests/smoketest.sh
RobotSail: That makes sense, let's keep it as is then.

tests/smoketest.sh
RobotSail: Yes, definitely something we can do in the future.

Maxusmusti (approved):

danmcp (approved):

PR #308, opened 2024-10-25
feat: add log exporting to e2e tests

### LINKED ISSUE #179: Include loss curve in E2E tests
To understand the impact that proposed changes may have, it's not enough for a training library to simply see functional tests pass. We need to also understand how said changes will affect the training quality and loss curve over time.

Therefore, we should add some GH automation that can take the exported loss data and render it as a matplotlib graph, overlaying it with the latest "main"/stable loss curve as a comparison. 

This way, we can have better confidence in our testing capabilities.

RobotSail:
Currently, the training library runs through a series of end-to-end tests which ensure there are
no bugs in the code being tested. However; we do not perform any form of validation to assure that
the training logic and quality has not diminished.

This presents an issue where we can potentially be "correct" in the sense of no hard errors being hit,
but invisible bugs may be introduced which cause models to regress in training quality, or other
bugs that plague the models themselves to seep in.

This commit fixes that problem by introducng the ability to export the training loss data itself
from the test and rendering the loss curve using matplotlib.

When the results are outputted, they can be found under the "Summary" tab of a Github actions run.
For example:

Screenshot 2024-10-25 at 6 18 14 PM

Resolves #179 

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -41,6 +41,11 @@ jobs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
     steps:
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
+        with:
+          egress-policy: audit
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
         with:
@@ -78,6 +83,11 @@ jobs:
     permissions: {}
 
     steps:
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
+        with:
+          egress-policy: audit
+
       - name: Install Packages
         run: |
           cat /etc/os-release
@@ -141,7 +151,22 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           . venv/bin/activate
-          ./scripts/e2e-ci.sh -m
+          # set preserve to true so we can retain the logs
+          ./scripts/e2e-ci.sh -mp
+          
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by `mktemp -d`. 
+          # Given this information, we can use the following command to find the file:
+          log_file=$(find /tmp -name "training_params_and_metrics_global0.jsonl")
+          mv "${log_file}" training-log.jsonl
+
+      - name: Upload training logs
+        uses: actions/upload-artifact@v4
+        with:
+          name: training-log.jsonl
+          path: ./instructlab/training-log.jsonl
+          retention-days: 1
+          overwrite: true
 
   stop-medium-ec2-runner:
     needs:
@@ -150,6 +175,11 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ always() }}
     steps:
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
+        with:
+          egress-policy: audit
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
         with:
@@ -164,6 +194,40 @@ jobs:
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           label: ${{ needs.start-medium-ec2-runner.outputs.label }}
           ec2-instance-id: ${{ needs.start-medium-ec2-runner.outputs.ec2-instance-id }}
+      
+      - name: Download loss data
+        id: download-logs
+        uses: actions/download-artifact@v4
+        with:
+          name: training-log.jsonl
+          path: downloaded-data
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-dev.txt
+      
+      - name: Try to upload to s3
+        id: upload-s3
+        continue-on-error: true
+        run: |
+          output_file='./test.md' 
+          python scripts/create-loss-graph.py  \
+            --log-file "${{ steps.download-logs.outputs.download-path }}/training-log.jsonl" \
+            --output-file "${output_file}" \
+            --aws-region "${{ vars.AWS_REGION }}" \
+            --bucket-name "${{ vars.AWS_S3_LOSS_GRAPHS_BUCKET_NAME }}" \
+            --base-branch "${{ github.event.pull_request.base.ref }}" \
+            --pr-number "${{ github.event.pull_request.number }}" \
+            --head-sha "${{ github.event.pull_request.head.sha }}" \
+            --origin-repository "${{ github.repository }}"
+
+          cat "${output_file}" >> "${GITHUB_STEP_SUMMARY}"
+
+      - name: Check S3 upload status
+        if: steps.upload-s3.outcome == 'failure'
+        run: |
+          echo "::warning::Failed to upload loss graph to S3. This won't block the workflow, but you may want to investigate."
+          echo "Loss graph upload failed" >> "${GITHUB_STEP_SUMMARY}"
 
   e2e-medium-workflow-complete:
     # we don't want to block PRs on failed EC2 cleanup
--- a/.github/workflows/e2e-nvidia-l40s-x4.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x4.yml
@@ -19,6 +19,12 @@ jobs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
+        with:
+          egress-policy: audit
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
         with:
@@ -54,6 +60,11 @@ jobs:
       pull-requests: write
 
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
+        with:
+          egress-policy: audit
       - name: Install Packages
         run: |
           cat /etc/os-release
@@ -170,7 +181,23 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           . venv/bin/activate
-          ./scripts/e2e-ci.sh -l
+
+          # set preserve to true so we can retain the logs
+          ./scripts/e2e-ci.sh -lp
+
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by `mktemp -d`. 
+          # Given this information, we can use the following command to find the file:
+          log_file=$(find /tmp -name "training_params_and_metrics_global0.jsonl")
+          mv "${log_file}" training-log.jsonl
+
+      - name: Upload training logs
+        uses: actions/upload-artifact@v4
+        with:
+          name: training-log.jsonl
+          path: ./instructlab/training-log.jsonl
+          retention-days: 1
+          overwrite: true
 
       - name: Add comment to PR if the workflow failed
         if: failure() && steps.check_pr.outputs.is_pr == 'true'
@@ -221,6 +248,12 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ always() }}
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
+        with:
+          egress-policy: audit
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
         with:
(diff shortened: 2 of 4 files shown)


.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -150,12 +165,14 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ always() }}
     steps:
+
```
danmcp, line 168: Was there a reason for adding the whitespace here and below?

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -164,6 +181,33 @@ jobs:
...
+          pip install -r requirements-dev.txt
+      
+      - name: Try to upload to s3
+        run: |
+          output_file='./test.md' 
+          python scripts/create-loss-graph.py  \
+            --log-file "${{ steps.download-logs.outputs.download-path }}/training-log.jsonl" \
+            --output-file "${output_file}" \
+            --aws-region "${{ vars.AWS_REGION }}" \
+            --bucket-name "${{ vars.AWS_S3_BUCKET_NAME }}" \
```
danmcp, line 203: This variable name should probably be a little more specific to the task:

Ex: AWS_S3_LOSS_GRAPHS_BUCKET_NAME

danmcp (changes requested):

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Yeah, probably an artifact from when I pasted over my changes into this file

RobotSail: Will delete

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Good point, I'll make that change

RobotSail: Done

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Done

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -156,6 +171,7 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ secrets.AWS_REGION }}
+
```
danmcp, line 174: Looks like there was one extra newline left here.  Was that one intentional?

RobotSail: No

JamesKunstle (approved):
lgtm!

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -141,7 +141,22 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           . venv/bin/activate
+          # set preserve to true so we can retain the logs
+          export PRESERVE=1
```
nathan-weinberg, line 145: Where is this var used?

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -141,7 +141,22 @@ jobs:
...
           ./scripts/e2e-ci.sh -m
+          
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by mktemp -d.
```
nathan-weinberg, line 149: small readability nit
```suggestion
          # and we know that it will be written into a directory created by `mktemp -d`
```

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -141,7 +141,22 @@ jobs:
...
+          export PRESERVE=1
           ./scripts/e2e-ci.sh -m
+          
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by mktemp -d. 
+          # Given this information, we can use the following command to find the file:
+          log_file=$(find /tmp -name "training_params_and_metrics_global0.jsonl")
+          mv "${log_file}" training-log.jsonl
+      - name: Upload training logs
+        uses: actions/upload-artifact@v4
```
nathan-weinberg, line 154: Can we use a hardened action here per the [org policy](https://github.com/instructlab/dev-docs/blob/main/docs/github-actions-use-policy.md#harden-the-github-action-runner)?

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Can you remove it?

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -164,6 +180,33 @@ jobs:
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           label: ${{ needs.start-medium-ec2-runner.outputs.label }}
           ec2-instance-id: ${{ needs.start-medium-ec2-runner.outputs.ec2-instance-id }}
+      
+      - name: Download loss data
+        id: download-logs
+        uses: actions/download-artifact@v4
```
nathan-weinberg, line 186: Same comment on hardening here

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -141,7 +141,22 @@ jobs:
...
+          # set preserve to true so we can retain the logs
+          export PRESERVE=1
           ./scripts/e2e-ci.sh -m
+          
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by mktemp -d. 
+          # Given this information, we can use the following command to find the file:
+          log_file=$(find /tmp -name "training_params_and_metrics_global0.jsonl")
+          mv "${log_file}" training-log.jsonl
+      - name: Upload training logs
```
nathan-weinberg, line 153:
```suggestion
          mv "${log_file}" training-log.jsonl

      - name: Upload training logs
```

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -164,6 +180,33 @@ jobs:
...
+      - name: Download loss data
+        id: download-logs
+        uses: actions/download-artifact@v4
+        with:
+          name: training-log.jsonl
+          path: downloaded-data
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-dev.txt
```
nathan-weinberg, line 193: Aren't dependencies already installed further up?

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -164,6 +180,33 @@ jobs:
...
+      
+      - name: Try to upload to s3
```
nathan-weinberg, line 195: So this says `Try to upload to s3` - do we assume a chance of failure? What is the job behavior if this step succeeds versus if it fails?

.github/workflows/e2e-nvidia-l40s-x4.yml
```diff
@@ -170,8 +170,25 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           . venv/bin/activate
+
+          # set preserve to true so we can retain the logs
+          export PRESERVE=1
           ./scripts/e2e-ci.sh -l
 
+          # we know that the file will be named something like f"/training_params_and_metrics_global{os.environ['RANK']}.jsonl" in python
+          # and we know that it will be written into a directory created by mktemp -d.
```
nathan-weinberg, line 179: Same comment as above

```suggestion
          # and we know that it will be written into a directory created by `mktemp -d`
```

.github/workflows/e2e-nvidia-l40s-x4.yml
```diff
@@ -235,3 +252,29 @@ jobs:
...
+      - name: Download loss data
+        id: download-logs
+        uses: actions/download-artifact@v4
+        with:
+          name: training-log.jsonl
+          path: downloaded-data
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-dev.txt
```
nathan-weinberg, line 265: Same comment as above

.github/workflows/e2e-nvidia-l40s-x4.yml
```diff
@@ -235,3 +252,29 @@ jobs:
...
+      
+      - name: Try to upload to s3
```
nathan-weinberg, line 267: Same comment as above

nathan-weinberg (changes requested):

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Inside of the e2e test script. It prevents the temp dir from being deleted.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Yes when I'm back from PTO

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Why not just use the flag?

RobotSail: @nathan-weinberg Could you please help me understand why we wouldn't want to use the env here? If we don't expect it to be used then we should just not have it in the script entirely.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: These little details are important so I'm glad you caught this.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Yes of course.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Added it 🫡

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: My understanding is that there are 3 jobs: 
- One to start the ec2 runner
- The ec2 runner
- One to stop the ec2 runner

Unless I'm mistaken, I believe this one has a separate scope to the ec2 runner, so it needs to install the dev deps again.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Should we fail the end-to-end run if we failed to upload to S3? If so, then we'd need to rerun the entire ec2 job. Else, we could succeed & provide a warning in the PR/summary that we could not upload to S3. 

There are two thoughts I'd have on this:
1. Testing for functional correctness is primary, loss correctness is secondary - here we consider the loss failing to upload as not a failure of the PR
2. Testing for loss + functional correctness is primary. Here we would hold a failure to upload to S3 with the same weight as themselves failing.

Let me know what your thoughts are on how we should proceed.

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Done

.github/workflows/e2e-nvidia-l40s-x4.yml
RobotSail: Answered in the other comment, basically isn't this running on another runner? In which case we'd need to reinstall dependencies.

.github/workflows/e2e-nvidia-l40s-x4.yml
RobotSail: Responded in the other thread - we need to decide whether failing to upload or not should null the results of an otherwise successful E2E run

.github/workflows/e2e-nvidia-l40s-x4.yml
RobotSail: I made the change, thank you for paying attention to details like this. It makes our codebase more enjoyable to work on in the long-term.

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: The flag just sets the env: https://github.com/instructlab/instructlab/blob/main/scripts/e2e-ci.sh#L433

I'm saying instead of setting it in the workflow file, just pass `-mp` instead of `-m`

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Thanks! Can you include the version number in a comment next to it like we do elsewhere?

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Was this done?

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Oh I see, so the deps are being installed here not on the AWS runner but on the GitHub runner orchestrating the overall action here?

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Honestly I'm fine with either, up to you and the @instructlab/training-maintainers! My suggestion: we go for 1 but include something like this? https://stackoverflow.com/questions/74907704/is-there-a-github-actions-warning-state

.github/workflows/e2e-nvidia-l40s-x4.yml
nathan-weinberg: Yeah let's have the convo in that thread so we don't trip over ourselves here :smile:

.github/workflows/e2e-nvidia-l40s-x4.yml
nathan-weinberg: Yup, same as the other let's just talk on that thread

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Yeah, the ec2 runner is only in charge of passing the log data back to the stop runner, but it does so through the use of github blobs (temporary GH S3-like storage).

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: Absolutely!

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: I see, that makes much more sense. I'll change it to use the flag then.

.github/workflows/e2e-nvidia-l4-x1.yml
nathan-weinberg: Gotcha, makes sense!

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: I've updated it to include this warning. Let me know if it lines up with what you expect.

RobotSail:
@nathan-weinberg I've updated the CI scripts with your feedback, please take another pass when you get a chance and make sure that we didn't miss anything.

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -41,6 +41,12 @@ jobs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
       ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
```
nathan-weinberg, line 46:
```suggestion
        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
```

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -78,6 +84,12 @@ jobs:
     permissions: {}
 
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
```
nathan-weinberg, line 89:
```suggestion
        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
```

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -150,6 +177,12 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ always() }}
     steps:
+      - name: "Harden Runner"
+        # v2.10.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7
```
nathan-weinberg, line 182:
```suggestion
        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
```

nathan-weinberg (approved):
I'd like the version number commenting to be consistent with how is it everywhere else, but otherwise LGTM

.github/workflows/e2e-nvidia-l4-x1.yml
RobotSail: I can make this change, I like this a lot more.

danmcp (approved):

RobotSail:
@nathan-weinberg This has been squashed, I'll remove the hold since that's the only issue.

PR #326, opened 2024-11-08
feat: add total_samples as a field to logs being emitted

### LINKED ISSUE #325: Log the total number of samples
Our goal for the training library is for other consumers to track information about the state of their training job.

In order to effectively track the progress of a training job, we need to emit the following information:

1. total number of samples per epoch
1. total number of epochs
1. samples seen in an epoch thus far

We already track #2, and #3, we just need to track #1.

The fix for this is really simple, we just need to to add a field to our log statement of `total_samples`.

An alternative to tracking this would be to instead enumerate over the dataloader and simply compute `step/len(train_loader)`.

RobotSail:
Today we emit a number of important values in the logs, but we do not explicitly announce how many samples
we are planning on iterating through. This commit emits total_samples as a field in both the initial log
as well as the subsequent logs during training. This way, this value can be accessed consistently by
any program interested in displaying information about its progress for convenience.

Fixes #325

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -18,6 +18,7 @@ from deepspeed.runtime.zero.utils import ZeRORuntimeException
 
 # pylint: disable=no-name-in-module
 from instructlab.dolomite.hf_models import GPTDolomiteForCausalLM
+from torch.utils.data import DataLoader
 from tqdm import tqdm
 from transformers import AutoModelForCausalLM, get_scheduler
 import torch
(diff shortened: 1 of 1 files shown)


jaideepr97 (approved):

JamesKunstle (approved):

JamesKunstle:
I approved then put the hold- quick q, is `len(dataloader.dataset)` a cached length value or is it computed each time? we could cache it before we start the loop if it's not cached already.

RobotSail:
@JamesKunstle What is your concern with the calculation of `len(dataloader.dataset)` if it's not cached? In Python, the `__len__` implementation should be $O(n)$. If this isn't the case then it would be a bug that needs to be fixed elsewhere.

JamesKunstle:
@RobotSail that tends to be the case for inbuilt stuff but this would be a length calculation for our custom Dataset objects. I confirmed that it's `O(1)` as best as I can tell so this should be good

PR #329, opened 2024-11-11
reorg main function; add 'hpu' option w/o implementation

JamesKunstle:
The `main` training function needed to be broken down into smaller
functions for readability/testability.

WORLD_SIZE, LOCAL_RANK, and RANK have also been extracted and made
global constants since they are set by the administrating
multiprocessing launcher (torchrun, in our case).

HPU configuration options and checks are also added.

Signed-off-by: James Kunstle <jkunstle@redhat.com>

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -10,6 +10,7 @@ import os
 import re
 import subprocess
 import time
+import typing
 
 # Third Party
 from accelerate import Accelerator
@@ -18,13 +19,15 @@ from deepspeed.runtime.zero.utils import ZeRORuntimeException
 
 # pylint: disable=no-name-in-module
 from instructlab.dolomite.hf_models import GPTDolomiteForCausalLM
+from torch.utils.data import DataLoader
 from tqdm import tqdm
 from transformers import AutoModelForCausalLM, get_scheduler
 import torch
 import torch.distributed
+import yaml
 
 # First Party
-from instructlab.training import config
+from instructlab.training import config, data_process
 from instructlab.training.async_logger import AsyncStructuredLogger
 from instructlab.training.config import (
     DataProcessArgs,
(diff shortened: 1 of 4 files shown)


src/instructlab/training/main_ds.py
```diff
@@ -512,106 +529,261 @@ def train(
...
+
+    init_distributed_training(
+        local_rank=LOCAL_RANK, world_size=WORLD_SIZE, hpu=args.hpu
+    )
+
+    flash_enabled = check_flash_attn_enabled(
+        disable_flash_attn=args.disable_flash_attn, use_dolomite=args.use_dolomite
+    )
+
+    tokenizer = configure_tokenizer(
```
JamesKunstle, line 741: I would have named this `setup_tokenizer` but that name would shadow the imported function.

src/instructlab/training/main_ds.py
```diff
@@ -512,106 +529,261 @@ def train(
...
+    """
 
-    args.samples_per_gpu = (
-        args.effective_batch_size // grad_accum // torch.distributed.get_world_size()
-    )
 
-    train_loader = setup_dataloader(
-        dataset,
-        tokenizer.pad_token_id,
+    this happens sometimes when we have more GPUs than data to process. In this case
```
JamesKunstle, line 633: will fix this docstring

src/instructlab/training/main_ds.py
```diff
@@ -512,106 +529,261 @@ def train(
...
+    optimizations that we can make) are ignored and we default to the `distributed data sampler`.
+    """
 
-    dataset = setup_dataset(
-        args.data_path,
-        mock=args.mock_data,
-        mock_len=args.mock_len,
-    )
+    # will try to make multipack work if possible.
+    sampler_type: str = "multipack"
```
RobotSail, line 594: `: str` here is redundant - it's inferred by the assignment

RobotSail (commented):
From a high level these changes look good, we'll just need to test them to verify.

PR #330, opened 2024-11-12
gaudi support training

JamesKunstle:
- **reorg main function; add 'hpu' option w/o implementation**
- **implement training for HPU**


JamesKunstle:
Builds on changes from #329. Still needs to be tested.

PR #336, opened 2024-11-13
feat: add toggle to pick legacy chat tmpl for granite

jaideepr97:
This PR adds a bool to train_args that allows consumers to specifically request the legacy ibm template if necessary for training

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -156,6 +156,9 @@ class TrainingArgs(BaseModel):
         os.path.dirname(__file__), "chat_templates/ibm_generic_tmpl.py"
     )
 
+    # this field determines if ibm_legacy_tmpl should be used instead
+    use_legacy_tmpl: bool = False
+
     # this field specifies the filepath to the training dataset before processing
     data_path: str
     ckpt_output_dir: str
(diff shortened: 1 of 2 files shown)


src/instructlab/training/config.py
```diff
@@ -156,6 +156,9 @@ class TrainingArgs(BaseModel):
         os.path.dirname(__file__), "chat_templates/ibm_generic_tmpl.py"
     )
 
+    # this field determins if ibm_legacy_tmpl should be used instead
```
alinaryan, line 159: nit: determins->determines

alinaryan (approved):

src/instructlab/training/config.py
```diff
@@ -156,6 +156,9 @@ class TrainingArgs(BaseModel):
...
+    use_legacy_sp_tokens: bool = False
```
Maxusmusti, line 160: maybe change to `use_legacy_tmpl`

src/instructlab/training/config.py
jaideepr97: fixed, thanks

src/instructlab/training/config.py
jaideepr97: fixed, thanks

Maxusmusti (approved):

PR #350, opened 2024-11-15
Update Dependencies to Move DeepSpeed to CUDA Extras

Maxusmusti:
Moving DeepSpeed requirement into CUDA extras, and making sure package installation is not required OOTB.

### DIFF
--- a/requirements-cuda.txt
+++ b/requirements-cuda.txt
@@ -1,5 +1,8 @@
 flash-attn>=2.4.0
 bitsandbytes>=0.43.1
 
+# available as an option for NVIDIA, FSDP still default
+deepspeed>=0.14.3
+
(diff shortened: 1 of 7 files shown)


Maxusmusti:
Tested that `pip install instructlab-training` -> `pip install instructlab-training[cuda]` path still works as expected, and that when DeepSpeed is not installed, default FSDP training path still works as expected.

Maxusmusti:
NOTE: Also capping Accelerate for now, as the recently released 1.1.x includes breaking changes that need to be resolved before it can be updated.

JamesKunstle (approved):

RobotSail (approved):
Let's Get This Money!!

PR #357, opened 2024-11-25
Updated token masking for new data "unmask" option (for pretraining samples)

Maxusmusti:
when `"unmask": true` is included in a data sample, ensures that non-system messages are all left unmasked (special tokens still masked)


JamesKunstle (approved):
lgtm!

PR #380, opened 2024-12-24
Adds support for knowledge distillation

RobotSail:
There are many different forms of model training which exist. One popular form of training is knowledge distillation, where a student model learns the output distributions from a teacher model.  This commit introduces support for knowledge distillation in the training library.

This commit also exposes the `weight_decay` hyperparameter which is often used to help deep learning models generalize.

Lastly, this commit changes the useage from `torch.distributed` to just `dist`, as it is a common module used throughout the codebase.

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -121,6 +121,16 @@ class DeepSpeedOptions(BaseModel):
     save_samples: int | None = None
 
 
+class DistillationConfig(BaseModel):
+    """
+    Config to use when performing knowledge distillation during training.
+    """
+
+    temperature: float = Field(1.0, gt=0.0)
+    alpha: float = Field(1.0, le=1.0, ge=0.0)
+    teacher_path: str
+
+
 # public API
 class ShardingStrategies(Enum):
     FULL_SHARD = "FULL_SHARD"
@@ -179,6 +189,11 @@ class TrainingArgs(BaseModel):
     is_padding_free: bool = False  # TODO: deprecate
     checkpoint_at_epoch: bool = True
     accelerate_full_state_at_epoch: bool = True
+    weight_decay: float = Field(0.0, ge=0.0)
+
+    # settings for knowledge distillation
+    distillation_options: Optional[DistillationConfig] = None
+    use_distillation: bool = False
 
     mock_data: Optional[bool] = False
     mock_data_len: int = 0
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -3,6 +3,7 @@
 # Standard
 from copy import deepcopy
 from pathlib import Path
+from typing import TYPE_CHECKING, Optional
 import argparse
 import json
 import math
@@ -40,9 +41,11 @@ except ImportError:
 from instructlab.dolomite.hf_models import GPTDolomiteForCausalLM
 from torch.utils.data import DataLoader
 from tqdm import tqdm
-from transformers import AutoModelForCausalLM, get_scheduler
+from transformers import AutoModelForCausalLM, PreTrainedTokenizer, get_scheduler
+from transformers.modeling_outputs import CausalLMOutput
 import torch
-import torch.distributed
+import torch.distributed as dist
+import torch.nn.functional as F
 
 # First Party
 from instructlab.training import config
@@ -71,6 +74,7 @@ from instructlab.training.utils import (
     create_lora_config,
     ensure_loadable_dolomite_checkpoint,
     load_latest_full_state,
+    log_rank_0,
     prepare_peft_model,
     prepare_universal_checkpoint_from_latest,
     retrieve_chat_template,
@@ -88,7 +92,7 @@ def setup_optimizer(args, model):
             model.parameters(),
             lr=args.learning_rate,
             betas=(0.9, 0.95),
-            weight_decay=0.0,
+            weight_decay=args.weight_decay,
         )
     elif args.distributed_training_framework == DistributedBackend.DEEPSPEED.value:
         # need to use this only when the CPU offload optimizer is enabled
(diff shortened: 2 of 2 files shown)


src/instructlab/training/config.py
```diff
@@ -121,6 +121,17 @@ class DeepSpeedOptions(BaseModel):
...
 
+# public API
+class DistillationConfig(BaseModel):
+    """
+    Config to use when performing knowledge distillation during training.
+    """
+
+    temperature: float = Field(1.0, gt=0.0)
+    alpha: float = Field(1.0, le=1.0, ge=0.0)
+    teacher_path: str
```
JamesKunstle, line 132: if possible would love to standardize on using pathlib.Path rather than str paths.

src/instructlab/training/main_ds.py
```diff
@@ -252,6 +260,38 @@ def make_inputs_require_grad(module, input, output):
...
+def setup_teacher_model(
+    model_name_or_path: str, device: torch.device, tokenizer: PreTrainedTokenizer
+):
+    """
+    Instantiates a teacher model to be used for distillation training.
+    """
+    teacher_model = AutoModelForCausalLM.from_pretrained(
+        model_name_or_path, torch_dtype=torch.bfloat16
+    ).to(device)
+    model_dev = next(teacher_model.parameters()).device
```
JamesKunstle, line 272: If you're calling `.to(device)` just above, could you make a note of why you also need to confirm the device locale below?

src/instructlab/training/config.py
```diff
@@ -179,6 +190,11 @@ class TrainingArgs(BaseModel):
     is_padding_free: bool = False  # TODO: deprecate
     checkpoint_at_epoch: bool = True
     accelerate_full_state_at_epoch: bool = True
+    weight_decay: float = Field(0.0, ge=0.0)
+
+    # settings for knowledge distillation
+    distillation_options: Optional[DistillationConfig] = None
```
JamesKunstle, line 196: I've seen that `Optional[DistillationConfig]` syntax is replaced by `DistillationConfig | None` in recent Pythonic parlance once the optional annotation was added to the language. This is a nit, not required to change.

src/instructlab/training/main_ds.py
```diff
@@ -382,11 +454,36 @@ def train(
...
+            output: CausalLMOutput = model(
                 **batch,
                 use_cache=False,
             )
-            loss = output.loss
+
+            loss = None
+            if args.distill:
+                # teacher_model should always be provided when `args.distill` is enabled
+                if TYPE_CHECKING:
```
JamesKunstle, line 467: I think this is supposed to be a runtime check but TYPE_CHECKING is always False at runtime.
https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

JamesKunstle: I think we should fail much earlier if distillation is set but no teacher_model is provided, like before we do any data preprocessing or fire up the GPUs.

src/instructlab/training/main_ds.py
```diff
@@ -382,11 +454,36 @@ def train(
...
+                    assert (
+                        teacher_model is not None
+                    ), "teacher model cannot be None when `distill` is enabled"
+
+                with torch.no_grad():
+                    teacher_output: CausalLMOutput = teacher_model(
```
JamesKunstle, line 473: You turn off `requires_grad` on all the params in the teacher_model. You could just be doing this instead, I think this gives the same output.

src/instructlab/training/main_ds.py
```diff
@@ -382,11 +454,36 @@ def train(
...
+                    student_output=output,
+                    teacher_output=teacher_output,
+                    alpha=args.distill_alpha,
+                    temp=args.distill_temp,
+                )
+
+            else:
+                loss = output.loss
+
+            assert loss is not None, "loss cannot be equal to None!"
```
JamesKunstle, line 486: `assert`s are typically not preferred in comparison to runtime exceptions.
```suggestion
            if loss is None:
                raise ValueError("loss was None during distillation training. Something unrecoverable went wrong.")
```

JamesKunstle: mostly because they can be removed with `-0` when the interpreter is invoked. But we want to check non-null all the time.

src/instructlab/training/config.py
RobotSail: @JamesKunstle I see your point, would it make sense for it to be a path when it can also take on a HF reference? I understand that references can technically still be paths, but to a consumer reading it might sound like only local models are accepted. Would `str | Path` be satisfactory?

src/instructlab/training/main_ds.py
RobotSail: Yes of course.

src/instructlab/training/config.py
RobotSail: Let's do the proposed method to be more consistent with how Python expects optionals in the future.

src/instructlab/training/main_ds.py
RobotSail: Sure I can change this. I was using them as scaffolding when writing this.

src/instructlab/training/main_ds.py
```diff
@@ -511,6 +609,9 @@ def main(args):
     # Third Party
     import yaml
 
+    if args.distill and not args.teacher_model_name_or_path:
```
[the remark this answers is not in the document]
RobotSail: Sweet :party-cat:

src/instructlab/training/main_ds.py
RobotSail: Yeah it is, I believe I had errors with type-checking here though and it not knowing that `teacher_model` is properly set.

src/instructlab/training/main_ds.py
RobotSail: I'm gonna make it not be specific to distillation training though since it's more about how we branch out. I suspect as we add other loss calculations (contrastive loss, preference tuning loss, etc.), we will start out by setting it to `None` and having this final check to ensure it was set to something.

PR #389, opened 2025-01-10
changes Fast unit CI runner, m8g->m7i

JamesKunstle:
m8g runner not available in us-east-2

Current CI broken because runner instance not available.

Signed-off-by: James Kunstle <jkunstle@redhat.com>

### DIFF
--- a/.github/workflows/unittesting-ci-nvidia.yaml
+++ b/.github/workflows/unittesting-ci-nvidia.yaml
@@ -19,7 +19,7 @@ on:
 
 env:
   pytest_mark: "fast"
-  ec2_runner_variant: "m8g.xlarge" # 4 Graviton CPU, 16GB RAM
+  ec2_runner_variant: "m7i.xlarge" # 4 Xeon CPU, 16GB RAM
 
 jobs:
   start-ec2-runner:


.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -19,7 +19,7 @@ on:
 
 env:
   pytest_mark: "fast"
-  ec2_runner_variant: "m8g.xlarge" # 4 Graviton CPU, 16GB RAM
+  ec2_runner_variant: "m7i.xlarge" # 4 Xeon CPU, 16GB RAM
```
danmcp, line 22: Since this is no longer on a GPU instance from the earlier iterations in the previous PR, what's the reason to not use a regular github runner?

JamesKunstle: I want the AMI that we're using on AWS- it's the same environment that I intend to run our other tests in, and it already has cuda installed. The tests that I'm planning to write for the 'fast' mark may still need cuda APIs in torch even if there are no devices present.

danmcp (approved):

RobotSail (approved):
LGTM

PR #390, opened 2025-01-10
chore: Change default temporary write directory in all e2e CI jobs from `tmpfs` to `/home/tmp`

courtneypacheco:
This PR tackles the same issue originally identified in https://github.com/instructlab/instructlab/issues/2866, where we noticed intermittent OOM issues during some of our e2e CI jobs.

While I have not noticed any OOM issues recently in this repo, updating the default temporary write directory from `tmpfs` (aka `/tmp`) to `/home/tmp` should mitigate the risk of OOM errors occurring in our e2e CI jobs since we'll now theoretically have access to 100% of the available physical memory in our EBS volume, instead of only 50%.

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -26,6 +26,7 @@ concurrency:
 
 env:
   LC_ALL: en_US.UTF-8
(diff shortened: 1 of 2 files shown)


nathan-weinberg (approved):

RobotSail (approved):
LGTM

PR #395, opened 2025-01-16
fix: Update ec2 instance type in unit test workflow file

### LINKED ISSUE #394: CI bug: Unit tests fail due to typo in git workflow
The unit tests in the CI immediately fail because a CI runner cannot be launched:

Image

This is because we have a typo in the unit testing workflow file that sets the instance type equal to the AWS region by mistake:

https://github.com/instructlab/training/blob/06879954b74485113f8fe868ebc022183df737d8/.github/workflows/unittesting-ci-nvidia.yaml#L51

courtneypacheco:
Resolves #394

Unit tests currently fail due to a minor configuration issue in one of our workflow files. This PR aims to resolve the typo. See referenced git issue.

### DIFF
--- a/.github/workflows/unittesting-ci-nvidia.yaml
+++ b/.github/workflows/unittesting-ci-nvidia.yaml
@@ -13,9 +13,11 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
   push:
+    # TEMPORARILY ADDING TESTING BRANCH. WILL REMOVE.
     branches:
       - "main"
       - "release-**"
+      - "fix-unit-test-workflow-file"
 
 env:
   pytest_mark: "fast"
@@ -26,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
-      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
 
     steps:
       - name: "Harden runner"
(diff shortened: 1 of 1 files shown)


.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -13,9 +13,11 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
   push:
+    # TEMPORARILY ADDING TESTING BRANCH. WILL REMOVE.
     branches:
       - "main"
       - "release-**"
+      - "fix-unit-test-workflow-file"
```
courtneypacheco, line 20: If you're good with the changes in this file, I will revert these temporary changes

.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -48,13 +50,15 @@ jobs:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ${{ vars.AWS_EC2_AMI }}
-          ec2-instance-type: ${{ vars.AWS_REGION }}
+          # TODO: Update the EC2 instance type to use the EC2 runner variant when GPU calls are mocked
+          # ec2-instance-type: ${{ env.ec2_runner_variant }}
+          ec2-instance-type: g4dn.12xlarge
```
danmcp, line 55: What's the logic for using this instance type?

.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -48,13 +50,15 @@ jobs:
...
           subnet-id: subnet-024298cefa3bedd61
           security-group-id: sg-06300447c4a5fbef3
           iam-role-name: instructlab-ci-runner
           aws-resource-tags: >
             [
-            {"Key": "Name", "Value": "instructlab-ci-github-large-runner"},
+            {"Key": "Name", "Value": "instructlab-ci-github-unit-test-runner"},
```
danmcp, line 61: This name wouldn't match the current pruning.  It's currently looking for `w+` between github and runner.  Either need to adjust here or in the pruning.

RobotSail (approved):

.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -128,10 +134,11 @@ jobs:
           aws-region: ${{ vars.AWS_REGION }}
 
       - name: "Stop EC2 runner"
```
danmcp, line 136: This doesn't appear to be setup so it will always run the way our other ec2-runner jobs are configured.

Ex: https://github.com/instructlab/instructlab/blob/main/.github/workflows/e2e-nvidia-l4-x1.yml#L150
Where it's a separate step that ways runs and depends on the previous step

JamesKunstle: I was wondering why that 'if(always())' step was in there. I'll add that to my PR #409

nathan-weinberg: Maybe be something leftover from an old job config - a lot of them have gotten out-of-sync, I had to make two hotfixes to the E2E Custom job in the Core repo today. @courtneypacheco has a Dev Doc open around consolidating some of these into common actions to make CI easier to maintain: https://github.com/instructlab/dev-docs/pull/179

JamesKunstle:
I incorporated these changes into #409 and added @courtneypacheco as the co-author on the commit with those changes. Thanks for your contribution Courtney!

PR #399, opened 2025-01-17
fixes unit ec2-name

JamesKunstle:
used the wrong environment variable for setting up the instance that we'd like to run our unit tests with.


danmcp (commented):
It's not clear on the intention of this PR vs https://github.com/instructlab/training/pull/395

JamesKunstle:
@danmcp I'm not sure what the proposed solution is in #395- this solution is to use the correct environment variable to specify the instance type to test with. I think they're trying to do the same thing, roughly.

JamesKunstle:
Replacing this PR with one that's based on a branch inside this repo:
#409

PR #404, opened 2025-01-21
adds upstream testing development document

JamesKunstle:
Adds a dev doc describing our testing strategy for this repo.

### DIFF
--- a/docs/upstream-testing-plan.md
+++ b/docs/upstream-testing-plan.md
@@ -1,0 +1,114 @@
+# Testing in 'instructlab/instructlab'
+
+We would like for this library to be a lightweight, efficient, and hackable toolkit for training LLMs with SFT and RLHF.
+
+Currently, this library is used heavily by the `ilab` tool from `github.com/instructlab/instructlab`. It is tested via that project's end-to-end tests.
+To improve development velocity, project trustworthiness, and library utility apart from `ilab`, we need to improve our project testing in-repo.
+
+## Testing a training library
+
+Training LLMs requires a lot of time and compute. We simply cannot "test" our library by executing an entire instruct-tuning run for each feature configuration we would like to verify.
+
+This presents a challenge: we have to test our code as well as possible without frequently generating a complete final artifact (a "trained" model checkpoint).
+
+To compensate for this challenge, we propose the following three-tier testing methodology:
+
+0. Linting: dev-time verification that code meets style and static analysis requirements
+1. Unit tests: verification that indivisible units work correctly
+2. Subsystem tests: verification that collections of units work correctly together
+3. Smoke tests: verification that feature-bounded systems work correctly
+4. Benchmark tests: verification that outputs behave correctly
+
+### Responsibilities of: Unit Tests
+
+This is one of the most common patterns in software testing and requires little introduction.
+For our purposes, Unit Tests will be tests that do not require accelerated hardware (e.g. GPUs).
+
+These tests will verify that the smallest units of meaningful code work correctly.
+They should be able to run on any development machine and should always run for incoming upstream PRs.
+
+### Responsibilities of: Subsystem Tests
+
+This might include data management utilities (distributed data sampler, batch collator, etc), model setup tools, loop instrumentation tools, etc.
+Subsystem Tests should give very fast confirmation that the library is behaving correctly, especially when bumping dependency versions.
+
+Subsystem Tests should be able to run on any development machine and should always run for incoming upstream PRs.
+
+### Responsibilities of: Smoke Tests
+
+The origin of the term "smoke test" comes from plumbing. Plumbers would pipe smoke, instead of water, through a completed pipe system to check that there were no leaks.
+In computer software, smoke testing provides fast verification to decide whether further testing is necessary- if something immediately breaks, we quickly know that there's a
+problem, and shouldn't spend resources on more in-depth testing.
+
+This is effectively what we'd like to do with this class of tests: verify that everything _seems_ to be in order before actually building a model to verify this.
+
+The challenge, however, is that "basic" testing in this context probably still requires accelerated hardware. We can't check that checkpointing a LoRA model while using CPU Offloading
+doesn't break without spinning up that exact testing scenario.
+
+"Smoke tests" in this repo, therefore, will be defined as tests that demonstrate the functionality, but don't prove the final model correctness, for features that we want to ship. This is
+very helpful to us because _most important failures don't happen during training itself- they happen during setup, checkpointing, and tear-down_.
+
+Pictorially, we can represent smoke testing as abbreviating a training run ('0' are setup/tear-down/checkpointing steps, '.' are training steps)"
+
+00000.................................................00000.....................................................00000
+
+to the following:
+
+00000.00000.00000
+
+Smoke tests would block PRs from being merged and would run automatically once all unit and lint tests pass.
+This puts the onus on the Unit Tests (and the test writer) to verify code correctness as much as they possibly can without using hardware. If something requires many invocations of
+smoke tests to debug, it probably wasn't sufficiently debugged during development, or is insufficiently unit tested.
+
+Smoke tests will inevitably require a high-spec development machine to run locally. It shouldn't be acceptable for smoke tests to run for >60 minutes- we should aim for them to run in <30 minutes.
+
+#### Smoke tests coverage
+
+This library attempts to provide training support for multiple variants of accelerators, numbers of accelerators, two distributed backends, and many features. Therefore, the completeness matrix
+is roughly explained by the following pseudocode:
+
+```python
+
+for variant in ["nvidia", "amd", "intel"]: # parallel at runner level
+    for framework in ["fsdp", "deepspeed"]:
+        for n_card in [1, 4]:
+            for feat_perm in all_permutations(["lora", "cpu_offload", None, ...]):
+
+                opts = [variant, n_card, framework, *feat_perm]
+
+                if check_opts_compatible(*opts):
+                    run_accelerated_tests(*opts)
+
+```
+
+When we have to add a new accelerator variant or feature, we can plug them into their respective tier in the smoke testing hierarchy.
+
+### Responsibilities of: Benchmark Tests
+
+Intermittently, we'll want to verify that our "presumeably correct" training library _actually produces_ a high-performance model.
+Predicting model performance from statically analyzing the logits or the parameters isn't viable, so it'll have to be benchmarked.
+
+Benchmark testing might look like the following:
+
+1. Tune a few models on well-understood dataset using one or two permutations of features, unique to benchmark run.
+2. Benchmark model on battery of evaluation metrics that give us the strongest signal re: model performance.
+
+Implementing this testing schema can be another project for a later time.
+
+## Concerns and Notes
+
+### Running smoke tests is expensive and shouldn't be done for every contribution
+
+Repository maintainers may want to implement a manual smoke-testing policy based on what a PR actually changes, rather than have the test be automatic. If an incoming PR changes the library interface but doesn't directly affect the training code, a unit test would probably be the most appropriate way to check whether all behavior still squares.
+
+### How often should we be running benchmark tests?
+
+There are many options for this depending on the observed risk that new features or dependency bumps seem to bring to the project. Likely, a comprehensive benchmark test should be done as a baseline and then the test should only be repeated when new model architectures are supported, major versions of ilab SDG are released, etc.
+
+### How should these tests be implemented? Bash scripts? Pytest?
+
+Pytest. Pytest + Tox. Writing tests in Bash is very hard to maintain.
+
+## Conclusion
+
+Testing ML models has never been easy, and it's unlikely that the final implementation of these ideas will closely mirror this document. However, we hope that the _spirit_ of this document- fast feedback, efficient use of resources, unit testing prioritization- will be effective.


nathan-weinberg:
@JamesKunstle can you fix the markdown linting? can run `make md-lint` for debugging locally

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+Smoke tests will inevitably require a high-spec development machine to run locally. It shouldn't be acceptable for smoke tests to run for >60 minutes- we should aim for them to run in <30 minutes.
+
+#### Smoke tests coverage
+
+This library attempts to provide training support for multiple variants of accelerators, numbers of accelerators, two distributed backends, and many features. Therefore, the completeness matrix
+is roughly explained by the following pseudocode:
+
+```python
+
+for variant in ["nvidia", "amd", "intel", "cpu", "mps"]: # parallel at runner level
```
danmcp, line 64: This is going to be a real challenge with runner availability to cover these on every PR.

JamesKunstle: Yeah, I figure that we'll scale out a little at a time, starting w/ CPU and nvidia, and adding runners to the matrix as possible.

JamesKunstle:
The current CI failure seems to be because the workflow can't access `vars.AWS_REGION` even though the workflow itself hasn't changed in this PR and the branch is rebased onto main, for which the unit test works. I'm not sure what that's happening.
@danmcp @courtneypacheco do ya'll have any guidance? I was assuming that the protected variables would be exposed to the workflows once they were merged.

danmcp:
> The current CI failure seems to be because the workflow can't access `vars.AWS_REGION` even though the workflow itself hasn't changed in this PR and the branch is rebased onto main, for which the unit test works. I'm not sure what that's happening. @danmcp @courtneypacheco do ya'll have any guidance? I was assuming that the protected variables would be exposed to the workflows once they were merged.

I am looking at the difference between:

https://github.com/instructlab/training/blob/main/.github/workflows/unit-tests.yaml#L12

and 

https://github.com/instructlab/training/blob/main/.github/workflows/e2e-nvidia-l4-x1.yml#L5

Were there reasons for the unit tests to be different?

JamesKunstle:
No there wasn't a reason for the two to be different apart from just writing the unit-test workflow from scratch. The two should have the same behavior, I'll amend that.

I'm confused about the `vars` context availability though because this CI workflow has merged to the main branch and doesn't seem to be able to access that context.

danmcp:
> No there wasn't a reason for the two to be different apart from just writing the unit-test workflow from scratch. The two should have the same behavior, I'll amend that.
> 
> I'm confused about the `vars` context availability though because this CI workflow has merged to the main branch and doesn't seem to be able to access that context.

pull_request runs from the context of your merge commit and pull_request_target runs from the context of instructlab/training

JamesKunstle:
That's super interesting, I totally missed that distinction in the docs. I'll look more closely at that.

I opened a PR #411 that mirrors the `on` and permissions usage between the e2e workflow invocation to the unit-tests invocation.

JamesKunstle:
From the docs:
"For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event."

My new understanding is that we have to use `pull_request_target` because we are setting up a self-hosted ec2 runner and need secrets / vars context to do this. We then limit the permissions of the job that runs the untrusted code.

If we weren't using these secure resources we could use the `pull_request` invocation option instead.

danmcp:
> My new understanding is that we have to use `pull_request_target` because we are setting up a self-hosted ec2 runner and need secrets / vars context to do this. We then limit the permissions of the job that runs the untrusted code.
> 
> If we weren't using these secure resources we could use the `pull_request` invocation option instead.

That matches my understanding as well.  I had originally made sure you had permissions set to {}, but I missed the pull_request vs pull_request_target at the top.

JamesKunstle:
Here's the most informative doc IMO: 
https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
+# Testing in 'instructlab/instructlab'
+
+We would like for this library to be a lightweight, efficient, and hackable toolkit for training "small" LLMs.
+Currently, this library is used heavily by `github.com/instructlab/instructlab` and the `ilab` tool, and is tested via that project's end-to-end tests.
```
booxter, line 4: > `github.com/instructlab/instructlab` and the `ilab` tool

Are these not the same?

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+1. Unit tests:              verification that subsystems work correctly
+2. Smoke tests:             verification that feature-bounded systems work correctly
+3. Benchmark tests:         verification that outputs behave correctly
+
+### Responsibilities of: Unit Tests
+
+This is one of the most common patterns in software testing and requires little introduction.
+For our purposes, Unit Tests will be tests that do not require accelerated hardware (e.g. GPUs).
+For us, the objective of a Unit Test is to check that an isolateable sub-system functions correctly.
```
booxter, line 22: A unit test, by definition, targets a unit. A (sub)system is, almost by definition, a collection of units (with their relationships).

What I'm saying is that we should have a separation of (1), e.g.:

(1) Unit tests (2) "(Sub)System" tests (3) Integration (multi-system) tests (4) Benchmarks.

In my opinion, instructlab project so far implemented most of its testing as 2+ or even 3+ only, and we have very few pure unit tests in the CLI repo, so perhaps this fact of life makes the distinction a bit fuzzy here.

(An example of a "not-really-a-unit" test in CLI repo would be any test that calls to functions of the CLI through click.Runner and observes the return code / output. These are not testing a unit - a class, a function... - but the whole machinery of CLI and should not have been called "unit", IMO.)

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+To improve development velocity, project trustworthiness, and library utility apart from `ilab`, we need to bolster our project testing.
+
+## Testing a training library
+
+Training LLMs requires a lot of time and compute. We simply cannot "test" our library by executing an entire instruct-tuning run for each feature configuration we would like to verify.
+This presents a challenge: we have to test our code as well as possible without frequently generating a complete final artifact (a "trained" model checkpoint).
+
+To compensate for this challenge, we propose the following three-tier testing methodology:
+
+1. Unit tests:              verification that subsystems work correctly
```
booxter, line 14: 0. Not strictly "testing" but a lot of real code issues - esp. in lack of more rigorous test coverage on other levelvs - are caught by linters / type checkers. I'd include them in testing methodology too.

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+The origin of the term "smoke test" comes from plumbing. Plumbers would pipe smoke, instead of water, through a completed pipe system to check that there were no leaks.
+In computer software, smoke testing provides fast verification to decide whether further testing is necessary- if something immediately breaks, we quickly know that there's a
+problem, and shouldn't spend resources on more in-depth testing.
+
+This is effectively what we'd like to do with this class of tests: verify that everything _seems_ to be in order before actually building a model to verify this.
+
+The challenge, however, is that "basic" testing in this context probably still requires accelerated hardware. We can't check that checkpointing a LoRA model while using CPU Offloading
+doesn't break without spinning up that exact testing scenario.
+
+"Smoke tests" in this repo, therefore, will be defined as tests that demonstrate the functionality, though not the correctness, of features that we want to ship. This is
```
booxter, line 40: > demonstrate the functionality, though not the correctness

I'm a bit confused by the distinction. I think smoke tests are checking for correctness, just for a more limited surface / scope. They are usually almost identical to the next step of the testing ladder, just using a smaller environment / running a lower number of test cases for the most basic scenarios.

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+For us, the objective of a Unit Test is to check that an isolateable sub-system functions correctly.
+
+This might include data management utilities (distributed data sampler, batch collator, etc), model setup tools, loop instrumentation tools, etc.
+Unit Tests should give very fast confirmation that the library is behaving correctly, especially when bumping dependency versions.
+
+Unit Tests should be able to run on any development machine and should always run for incoming upstream PRs.
+
+### Responsibilities of: Smoke Tests
+
+The origin of the term "smoke test" comes from plumbing. Plumbers would pipe smoke, instead of water, through a completed pipe system to check that there were no leaks.
```
cdoern, line 31: today I learned

docs/upstream-testing-plan.md
cdoern: comment here: there are no MPS runners to my knowledge. We would need to self host runners for most of these scenarios outside of CPU and CUDA

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Concerns and Notes
+
+### Running smoke tests is expensive and shouldn't be done for every contribution
+
+Repository maintainers may want to implement a manual smoke-testing policy based on what a PR actually changes, rather than have the test be automatic. If an incoming PR changes the library interface but doesn't directly affect the training code, a unit test would probably be the most appropriate way to check whether all behavior still squares.
+
+### How often should we be running benchmark tests?
+
+There are many options for this depending on the observed risk that new features or dependency bumps seem to bring to the project. Likely, a comprehensive benchmark test should be done as a baseline and then the test should only be repeated when new model architectures are supported, major versions of ilab SDG are released, etc.
```
cdoern, line 98: we might want to do this more often, like on Medium-Large library changes. Or upon a weekly release to ensure we didn't merge anything in the last week that broke stuff

cdoern (approved):
this looks good, few comments

docs/upstream-testing-plan.md
booxter: I see `O(n^3)` configs here. This is indeed not scalable. In some projects with similar problems I worked for, this dilemma was solved by crafting a set of "verified configurations" that were defined in such a way so that they cover the highest number of unique paths through the multi-dimentional feature space without blowing up the number of permutations so much.

For example, one could have:

1. NVIDIA + FSDP + 8 cards
2. AMD + DEEPSPEED + 2 cards
3. INTEL + DEEPSPEED + 1 card
4. CPU + "NONE" + No cards
5. MPS + "NONE" + No cards (or should it be 1 card if we can find a runner with MPS properly working?)

The list can then be expanded as needed if we feel like we want to have better coverage for some feature paths (f.e. test NVIDIA with both frameworks and in two card configuration - 1 vs 8).

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Responsibilities of: Benchmark Tests
+
+Intermittently, we'll want to verify that our "presumeably correct" training library _actually produces_ a high-performance model.
+Predicting model performance from statically analyzing the logits or the parameters isn't viable, so it'll have to be benchmarked.
+
+Benchmark testing might look like the following:
+
+1. Tune a few models on well-understood dataset using one or two permutations of features, unique to benchmark run.
+2. Benchmark model on battery of evaluation metrics that give us the strongest signal re: model performance.
```
booxter, line 86: One question here is whether instructlab/training library should depend on instructlab/eval library for its own benchmarking, or whether a more generic library (lm-eval) should be used instead. Does eval library bring anything unique for testing purposes?

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+
+To compensate for this challenge, we propose the following three-tier testing methodology:
+
+1. Unit tests:              verification that subsystems work correctly
+2. Smoke tests:             verification that feature-bounded systems work correctly
+3. Benchmark tests:         verification that outputs behave correctly
+
+### Responsibilities of: Unit Tests
+
+This is one of the most common patterns in software testing and requires little introduction.
```
booxter, line 20: I'd include automated coverage targets / reports here. I know coverage targets are a bit controversial (Goodhart's law), but assuming we understand the risks, coverage reports and enforcement should help us not to take sight off the ball.

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,102 @@
...
+smoke tests to debug, it probably wasn't sufficiently debugged during development, or is insufficiently unit tested.
+
+Smoke tests will inevitably require a high-spec development machine to run locally. It shouldn't be acceptable for smoke tests to run for >60 minutes- we should aim for them to run in <30 minutes.
+
+#### Smoke tests coverage
+
+This library attempts to provide training support for multiple variants of accelerators, numbers of accelerators, two distributed backends, and many features. Therefore, the completeness matrix
+is roughly explained by the following pseudocode:
+
+```python
```
booxter, line 62: I'd recommend to explicitly state that tests are written in python using pytest framework. I'd rather not perpetuate the "functional tests" bash shell pattern found elsewhere in this repo, if possible.

docs/upstream-testing-plan.md
JamesKunstle: Yeah agreed, testing the _entire_ space is intractable.

I think we can probably restrict the space to:
[variants] x [fsdp or deepspeed] x [single or multi-card]

We can assume that the topology is >=1 accelerator and >=1 node. We shouldn't be concerned with CPU or MPS right now.

docs/upstream-testing-plan.md
JamesKunstle: Yeah this could be worded better.
Essentially what I'm describing is that we'd be looking for "loss went down" in the test, rather than training until convergence and benchmarking.

docs/upstream-testing-plan.md
JamesKunstle: You're right, fixed.

docs/upstream-testing-plan.md
JamesKunstle: added that

docs/upstream-testing-plan.md
JamesKunstle: That's a good idea. We can make an issue to investigate this in the future

docs/upstream-testing-plan.md
JamesKunstle: I don't have a clear answer on this upfront. I feel like we'd want to dogfood our own lib but be flexible

booxter (approved):

ktdreyer (approved):

RobotSail (changes requested):
Thank you @JamesKunstle for providing this document, having an automated way of validating incoming code will be very valuable to have in this library. 

After reviewing this document, I'm requesting these changes

- **Benchmarks**:  Measuring training library correctness through benchmarks will be unreliable for testing a *training library* and will require us to run many benchmarks to have a correct result. For example, a model training incorrectly on AMD might score well on MMLU but will also recommend users to drink and drive. Additionally, it will be very computationally expensive to run these so the ROI here would be very low
- **Plotting loss curves**: An easy way to test the correctness for something complex like a training library would be to create a system in our testing where we can export and store loss data from all of the training runs we do, and then have a simple way to compare it against the matrix of configurations we've included here, and past runs we've done historically

This way we can make testing into something viable in the training library, without greatly impacting friction of adding contributions

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,114 @@
+# Testing in 'instructlab/instructlab'
+
+We would like for this library to be a lightweight, efficient, and hackable toolkit for training LLMs with SFT and RLHF.
```
RobotSail, line 3: What is meant by RLHF in this case?

JamesKunstle: "Reinforcement Learning from Human Feedback"

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,114 @@
...
+
+to the following:
+
+00000.00000.00000
+
+Smoke tests would block PRs from being merged and would run automatically once all unit and lint tests pass.
+This puts the onus on the Unit Tests (and the test writer) to verify code correctness as much as they possibly can without using hardware. If something requires many invocations of
+smoke tests to debug, it probably wasn't sufficiently debugged during development, or is insufficiently unit tested.
+
+Smoke tests will inevitably require a high-spec development machine to run locally. It shouldn't be acceptable for smoke tests to run for >60 minutes- we should aim for them to run in <30 minutes.
```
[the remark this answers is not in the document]
JamesKunstle: I'd like to keep the upper-bound to 30min. That's the hard cutoff- faster is always better.

docs/upstream-testing-plan.md
```diff
@@ -0,0 +1,114 @@
...
+                opts = [variant, n_card, framework, *feat_perm]
+
+                if check_opts_compatible(*opts):
+                    run_accelerated_tests(*opts)
+
+```
+
+When we have to add a new accelerator variant or feature, we can plug them into their respective tier in the smoke testing hierarchy.
+
+### Responsibilities of: Benchmark Tests
```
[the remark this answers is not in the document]
JamesKunstle: Yeah likely true. I can remove this / deprioritize it.

gkneighb:
Took a careful read of the doc and the full review history. The PR has 3 approvals and is blocked on @RobotSail's CHANGES_REQUESTED — two threads are still unresolved on the current head:

1. **Smoke-test runtime budget (L63).** @RobotSail proposed a 5-minute ceiling; the original text was 60-min ceiling / 30-min target; you pushed back wanting 30 minutes. A workable middle: **30-minute hard cap per matrix entry, target under 10 minutes**, with explicit guidance that consistently-over-cap entries get split, parallelized, or promoted to the slower tier. Captures the "fast feedback" spirit without putting an unrealistic ceiling on multi-config matrices.

2. **Benchmark tier (L86).** You agreed on 2025-04-23 to deprioritize but didn't push the edit. Concrete proposal: **rename the tier to "Convergence Checks"** and reframe it as loss-curve comparison against a stored baseline — short fixed-seed runs per smoke-matrix entry, baseline refreshed deliberately on new arch / major dep bumps / training-loop changes. Downstream benchmarks (MMLU, MT-Bench) move out of the per-PR tiers and into release qualification, with the `instructlab/eval` vs `lm-eval-harness` choice called out as deferred (picks up @booxter's L94 question).

A few smaller things I caught while reading:

- The H1 still says `# Testing in 'instructlab/instructlab'`; this doc lives in `instructlab/training`.
- The smoke-test ASCII diagram (lines ~53, 57) isn't in a code fence, so markdown collapses the whitespace it relies on.
- The bullet list at the top includes `0. Linting` but the body has no matching section.
- The `all_permutations([..., None, ...])` pseudocode is O(n!) and includes a confusing `None`-as-feature element. @booxter's verified-configurations idea — and your reply agreeing the topology should be `>=1 accelerator, no CPU/MPS` — fits much better as a hand-curated list of named configs.

I've staged all of the above as a 5-commit follow-up on top of your branch (lint passes). Happy to either (a) open it as a separate PR that supersedes this one if you've moved on, or (b) push to your branch if you'd prefer to keep your authorship and just pick the commits you agree with. Let me know which you prefer — and if you want to push back on any of the resolutions above, that's fine too, the goal is just to unstick the PR.

gkneighb:
Saw this got closed — no worries either way, but wanted to check whether the team's decided to abandon this dev doc, replace it with a different approach, or whether there's a path that would make the edits useful. Either way, happy to redirect effort. Thanks!

RobotSail:
Hi @gkneighb , thank you for taking interest in this issue and reviewing the PR! This PR has been abandoned and is no longer reflects the current project's scope. 

If you are interested in contributing to the testing framework for this repo, I recommend raising a new issue so we can continue discussion there.

Also, if you haven't checked it out yet, you should look at the [training hub](https://github.com/Red-Hat-AI-Innovation-Team/training_hub) repo to understand how this library gets consumed. Our current goal is to provide an easy-to-use and accessible interface for common training algorithms. We consume instructlab/training as the primary backend for SFT.

PR #409, opened 2025-01-23
fixes unit ec2-name

JamesKunstle:
Recreating the PR from a branch that has access to repo environment variables + secrets.

### DIFF
--- a/.github/workflows/unittesting-ci-nvidia.yaml
+++ b/.github/workflows/unittesting-ci-nvidia.yaml
@@ -19,14 +19,15 @@ on:
 
 env:
   pytest_mark: "fast"
-  ec2_runner_variant: "m7i.xlarge" # 4 Xeon CPU, 16GB RAM
+  ec2_runner_variant: "g4dn.12xlarge" # T4 machine that would support an nvidia-smi call.
+  # ec2_runner_variant: "m7i.xlarge" # 4 Xeon CPU, 16GB RAM
 
 jobs:
   start-ec2-runner:
     runs-on: ubuntu-latest
     outputs:
       label: ${{ steps.start-ec2-runner.outputs.label }}
-      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id}}
 
     steps:
       - name: "Harden runner"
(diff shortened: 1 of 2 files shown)


.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ${{ vars.AWS_EC2_AMI }}
-          ec2-instance-type: ${{ vars.AWS_REGION }}
+          ec2-instance-type: ${{ env.ec2_runner_variant }}
           subnet-id: subnet-024298cefa3bedd61
           security-group-id: sg-06300447c4a5fbef3
           iam-role-name: instructlab-ci-runner
```
danmcp, line 54: I know you haven't changed it in this pr.  But we should change instructlab-ci-github-large-runner to something like:

instructlab-ci-github-unittest-runner 

I remembered after the initial pull that large will imply a much longer max runtime.

JamesKunstle: Yes agreed. What will need to change in AWS logic for this?

danmcp: > Yes agreed. What will need to change in AWS logic for this?

Nothing it will fit into the general ci runner bucket with that name.

.github/workflows/unittesting-ci-nvidia.yaml
```diff
@@ -19,14 +19,15 @@ on:
 
 env:
   pytest_mark: "fast"
-  ec2_runner_variant: "m7i.xlarge" # 4 Xeon CPU, 16GB RAM
+  ec2_runner_variant: "g4dn.12xlarge" # T4 machine that would support an nvidia-smi call.
```
danmcp, line 22: What's the reason for the 12xlarge?

JamesKunstle: Initially I was curious about the same thing. I don't think we need the 12xlarge, just some machine that's GPU-enabled if we're going to be checking on nvidia-smi. That being said, I think these unit tests need to be isolated from an accelerator so I'll make that change too. the `smoke test` variants will need hardware, the unit tests won't.

JamesKunstle:
@courtneypacheco I learned from your PR [here](https://github.com/instructlab/training/pull/395) to get this test to pass. Thanks for identifying the changes that needed to be made! I'd like to merge this PR when I've changed the instance type and removed the nvidia-smi call. Would that be okay with you? I think that would close your PR w/ you further work required on your end.

danmcp:
> @courtneypacheco I learned from your PR [here](https://github.com/instructlab/training/pull/395) to get this test to pass. Thanks for identifying the changes that needed to be made! I'd like to merge this PR when I've changed the instance type and removed the nvidia-smi call. Would that be okay with you? I think that would close your PR w/ you further work required on your end.

@JamesKunstle You should add @courtneypacheco as a Co-author on your commit

JamesKunstle:
> > @courtneypacheco I learned from your PR [here](https://github.com/instructlab/training/pull/395) to get this test to pass. Thanks for identifying the changes that needed to be made! I'd like to merge this PR when I've changed the instance type and removed the nvidia-smi call. Would that be okay with you? I think that would close your PR w/ you further work required on your end.
> 
> @JamesKunstle You should add @courtneypacheco as a Co-author on your commit

Definitely- just updated.

.github/workflows/unittesting-ci-nvidia.yaml
JamesKunstle: OK done

danmcp (approved):

PR #416, opened 2025-01-27
Using AutoConfig to load model config file.

abhi1092:
* Right now loading from local file causes the code to fail on HF checkpoints.
* This  should fix that issue


Maxusmusti (approved):
LGTM

RobotSail (approved):

PR #422, opened 2025-02-01
Disable medium test by default with older release branches

danmcp:
Avoids more complicated changes to avoid using HF_TOKEN.


bbrowning (approved):
Approving this one as well, since we're making this change across all repos.

PR #424, opened 2025-02-04
Adds smoke test workflow and tests

JamesKunstle:
adds a smoke-test.yaml workflow that's very similar to unit-test.yaml but calls a different tox+pytest path and requires a cuda runner.


JamesKunstle:
This PR removes the reusable workflow contributions in https://github.com/instructlab/training/pull/419 but keeps the smoke testing.

JamesKunstle:
This PR needs to go in so that we can test the manual functionality of the smoke-test addition. It works on a dev machine.

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+
+# matrix of training environments we'd like to test
+DIST_BACKEND_FRAMEWORKS = ["fsdp", "deepspeed"]
+USE_DOLOMITE = [True, False]
+CPU_OFFLOADING = [True, False]
+USE_LORA = [True, False]
+
+
+@pytest.fixture(scope="module")
+def custom_tmp_path():
```
ktdreyer, line 59: (Apologies for the drive-by comment. The rest of this PR is over my head :laughing: . I was just reading this PR and noticed something related to this fixture.)

The implementation of this method looks similar to [pytest's own `tmp_path` fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html). I wondered if you could call `custom_tmp_path(tmp_path)` and simply return `tmp_path` here, or even use `tmp_path_factory` if you want a session-scoped fixture.

JamesKunstle: It's been a while since I wrote this- I kinda remember _doing_ exactly what you recommend and having the behavior be strange, and this was a convenient workaround. I don't really remember _why_ though- I obviously should have made a note!

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -156,7 +157,7 @@ jobs:
           . venv/bin/activate
           # set preserve to true so we can retain the logs
           ./scripts/e2e-ci.sh -mp
-          
+
           # HACK(osilkin): The above test runs the medium workflow test which does not actually test the training library.
```
cdoern, line 161: so, if this job doesn't use the training library should we remove it?

cdoern: this uses `--pipeline full` which uses the full loop from ilab (sorry I did this 😆 )

cdoern: this might be tangential to this PR but might be nice to see a green CI on this by just removing the test.

.github/workflows/smoke-tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+name: "Run smoke tests via Tox::pytest"
+# These tests will be long running and require accelerated hardware.
+# They will help to verify that the library is *functionally* correct but
+# will not try to verify that the libary is *correct*.
+
+on:
+  # TEMP - only runs when manually invoked
+  # and only runs against branches in the repo.
+  workflow_dispatch:
```
cdoern, line 11: you can add on_pull so that it runs and passes on this PR!

requirements-dev.txt
```diff
@@ -13,3 +13,4 @@ ipython
 ipykernel
 jupyter
 
+huggingface_hub
```
cdoern, line 16: do we want a version for that?

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+from instructlab.training.config import (
+    DataProcessArgs,
+    DistributedBackend,
+    TorchrunArgs,
+    TrainingArgs,
+)
+from instructlab.training.main_ds import run_training
+
+MINIMAL_TRAINING_ARGS = {
+    "max_seq_len": 140,  # this config fits nicely on 4xL40s and may need modification for other setups
```
cdoern, line 24: this makes sense, _but_ if we want to use these PROFILES then we should scope work to lift and shift the system profiles from ilab to the training repo

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+    ckpt_dir = custom_tmp_path / "checkpoints"
+    ckpt_dir.mkdir()
+
+    yield ckpt_dir
+
+    shutil.rmtree(ckpt_dir)
+
+
+@pytest.fixture(scope="module")
+def prepared_data_dir(custom_tmp_path: pathlib.Path) -> pathlib.Path:
```
cdoern, line 83: missing docstring

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+
+    return cached_model_dir
+
+
+def this_file_path() -> pathlib.Path:
+    return pathlib.Path(__file__).resolve()
+
+
+def data_in_repo_path() -> pathlib.Path:
```
cdoern, line 125: missing docstring

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+    current_file_path = this_file_path()
+    data_in_repo_path = (
+        current_file_path.parents[2] / "sample-data" / "train_all_pruned_SDG.jsonl"
+    )
+    return data_in_repo_path
+
+
+def chat_template_in_repo_path() -> pathlib.Path:
```
cdoern, line 133: docstring plz

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+        **MINIMAL_TRAINING_ARGS,
+    )
+
+    torch_args = TorchrunArgs(**DEFAULT_TORCHRUN_ARGS)
+
+    run_training(torch_args=torch_args, train_args=train_args)
+    assert True
+
+
+@pytest.mark.slow
```
cdoern, line 197: so we aren't skipping this one but its also running training, why is that?

cdoern (commented):
initial pass, some of these are nits, some are comments for the future. Overall I like the shape of this

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+    "rdzv_id": 123,
+    "rdzv_endpoint": "127.0.0.1:12345",
+}
+
+REFERENCE_TEST_MODEL = "instructlab/granite-7b-lab"
+RUNNER_CPUS_EXPECTED = 4
+
+# matrix of training environments we'd like to test
+DIST_BACKEND_FRAMEWORKS = ["fsdp", "deepspeed"]
+USE_DOLOMITE = [True, False]
```
booxter, line 53: remove for now those that are not used.

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+
+    torch_args = TorchrunArgs(**DEFAULT_TORCHRUN_ARGS)
+
+    run_training(torch_args=torch_args, train_args=train_args)
+    assert True
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("dist_backend", DIST_BACKEND_FRAMEWORKS)
+@pytest.mark.parametrize("cpu_offload", CPU_OFFLOADING)
```
booxter, line 199: I'd rather inline the true / false selector here.

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+
+    torch_args = TorchrunArgs(**DEFAULT_TORCHRUN_ARGS)
+
+    run_training(torch_args=torch_args, train_args=train_args)
+    assert True
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("dist_backend", DIST_BACKEND_FRAMEWORKS)
```
booxter, line 198: please calculate values from DistributedBackend.FSDP enum here.

```
[backend.value for backend in (DistributedBackend.FSDP, DistributedBackend.XXX)]
```

tox.ini
```diff
@@ -19,14 +19,35 @@ basepython = python3.11
...
 description = run unit tests with pytest
-commands = {envpython} -m pytest tests {posargs}
+passenv =
+	HF_HOME
+deps = 
+    pytest
+    pytest-asyncio
+    pytest-cov
+    pytest-html
+    -r requirements-dev.txt
```
booxter, line 29: Should some of these go to requirements-dev.txt itself? (what's the distinction between putting dependencies here and in that requirements file?)

tox.ini
```diff
@@ -19,14 +19,35 @@ basepython = python3.11
 
 [testenv:py3-unit]
 description = run unit tests with pytest
-commands = {envpython} -m pytest tests {posargs}
+passenv =
+	HF_HOME
+deps = 
+    pytest
+    pytest-asyncio
+    pytest-cov
```
booxter, line 27: do we use cov / html anywhere?

tox.ini
```diff
@@ -19,14 +19,35 @@ basepython = python3.11
...
+    pytest-html
+    -r requirements-dev.txt
+commands = {envpython} -m pytest tests/test_unit {posargs}
 # NOTE: {posargs} is a placeholder for input positional arguments
 # such as `tox -e py3-unit -- --pdb` if we wanted to run pytest with pdb enabled.
 # `--` delimits flags that are meant for tox vs. those that are positional arguments for
 # the command that's being run in the environment.
 
 # format, check, and linting targets don't build and install the project to
```
booxter, line 36: this comment belongs to the next section

tox.ini
```diff
@@ -19,14 +19,35 @@ basepython = python3.11
...
 # speed up testing.
+[testenv:py3-smoke]
+description = run accelerated smoke tests with pytest
+passenv =
+	HF_HOME
```
booxter, line 41: you could define it once in `testenv` section not to repeat

tox.ini
```diff
@@ -19,14 +19,35 @@ basepython = python3.11
...
+deps = 
+    pytest
+    pytest-asyncio
+    pytest-cov
+    pytest-html
+    -r requirements-dev.txt
+    -r requirements-cuda.txt
```
booxter, line 48: what if I want to run with a different accelerator? is it a choice to be made by the code that sets the test environment up? (github workflow?)

booxter:
I'd like to see @courtneypacheco or @ktdreyer to confirm the AWS runner configuration here is valid. And if any of this actions code could be shifted to common actions defined for the whole org.

.github/workflows/e2e-nvidia-l4-x1.yml
JamesKunstle: Should we do that in a follow-up so we're doing on thing at a time in this PR?

cdoern: sure!

.github/workflows/smoke-tests.yaml
```diff
@@ -0,0 +1,143 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: "Run smoke tests via Tox::pytest"
+# These tests will be long running and require accelerated hardware.
+# They will help to verify that the library is *functionally* correct but
+# will not try to verify that the libary is *correct*.
```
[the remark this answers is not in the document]
JamesKunstle: True- I was trying to express that this will run tests that _execute_ the code but won't benchmark the output model to see if _it's_ better. That'd be my definition of a correctness test for this.

.github/workflows/smoke-tests.yaml
JamesKunstle: true- I should do that

.github/workflows/smoke-tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: "Install packages"
+        run: |
+          cat /etc/os-release
+          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
+
+      - name: "Verify cuda environment is setup"
+        run: |
+          export CUDA_HOME="/usr/local/cuda"
+          export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64"
```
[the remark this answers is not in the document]
JamesKunstle: gotcha, updated and changed the bash variable expansion to be correctly scoped

tests/test_smoke/test_train.py
JamesKunstle: true, let's scope that for future work

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+
+    temp_path = pathlib.Path(temp_dir)
+
+    yield temp_path
+
+    shutil.rmtree(temp_path)
+
+
+@pytest.fixture(scope="function")
+def checkpoint_dir(custom_tmp_path: pathlib.Path) -> pathlib.Path:
```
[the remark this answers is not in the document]
JamesKunstle: You're right! I didn't know how to type that, this makes a lot more sense

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+
+    shutil.rmtree(ckpt_dir)
+
+
+@pytest.fixture(scope="module")
+def prepared_data_dir(custom_tmp_path: pathlib.Path) -> pathlib.Path:
+    data_file_dir = custom_tmp_path / "prepared_data"
+    data_file_dir.mkdir()
+
+    return data_file_dir
```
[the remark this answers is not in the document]
JamesKunstle: I think I only need that pattern twice- once for the temporary storage and another time for the checkpoints dir.
Otherwise, we're just creating the dirs for the model and the data.

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+        / "chat_templates"
+        / "ibm_generic_tmpl.py"
+    )
+    return chat_template_path
+
+
+# TODO: This uses our data preprocessing utility which is not, itself, well tested.
+# need to write tests for this as well.
+@pytest.fixture(scope="module")
+def cached_training_data(
```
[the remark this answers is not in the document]
JamesKunstle: true- when we have more tests we can refactor slightly. I'd like to start in a single file just for the moment

tests/test_smoke/test_train.py
```diff
@@ -0,0 +1,227 @@
...
+@pytest.mark.skip
+@pytest.mark.slow
+def test_basic_training_run(
+    cached_test_model: pathlib.Path,
+    cached_training_data: pathlib.Path,
+    checkpoint_dir: pathlib.Path,
+    prepared_data_dir: pathlib.Path,
+) -> None:
+    """
+    Used for isolated test development. Skipped when not in use.
```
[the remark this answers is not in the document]
JamesKunstle: this was a development artifact. removed.

tox.ini
JamesKunstle: I'm isolating this to the smoketest case specifically- I'm thinking that if it needs to be more broadly available we can change it as you say

tox.ini
JamesKunstle: Yeah I think the workflow has to call specific code. This is set to cuda because we only have cuda runners

.github/workflows/smoke.yaml
```diff
@@ -0,0 +1,148 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: "Run smoke tests via Tox::pytest"
+# These tests will be long running and require accelerated hardware.
+
+on:
+  # TEMP - only runs when manually invoked
```
booxter, line 7: is it? isn't pull_request_target now triggering it automatically?

tox.ini
booxter: fair but should you then passenv for unit tests like it's done here?

.github/workflows/smoke.yaml
courtneypacheco: We could definitely update this comment to be more accurate, but I do think it could actually be valuable to have `workflow_dispatch` in addition to `pull_request_target`.  If you have `workflow_dispatch` set, then you can trigger these smoke tests any time on existing branches. This could be particularly useful if you want to quickly rerun the smoke tests against  a release branch when you know there has been a CUDA update.

.github/workflows/smoke.yaml
```diff
@@ -0,0 +1,148 @@
...
+    steps:
+      - name: "Harden runner"
+        uses: step-security/harden-runner@cb605e52c26070c328afc4562f0b4ada7618a84e # v2.10.1
+        with:
+          egress-policy: audit
+
+      - name: "Install packages"
+        run: |
+          cat /etc/os-release
+          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
```
ktdreyer, line 87: I've been trying to clean this up in `instructlab` itself (https://github.com/instructlab/instructlab/pull/3140), so when I see it here, we can also simplify it:
```suggestion
          sudo dnf install -y gcc gcc-c++ make git-core python3.11 python3.11-devel
```

ktdreyer (approved):

booxter (approved):

.github/workflows/smoke.yaml
JamesKunstle: I'll update the comment

PR #428, opened 2025-03-05
Updates data processing logic to remove dependency on hardcoded chat templates

RobotSail:
In order to use the training library today, we need to manually define the chat template and rely on specifically parsing the special tokens.

This introduces a number of issues for consumers where other models cannot easily be used without some inital effort.

This PR resolves this issue by introducing a new form of data processing which is able to arbitrarily process data in the messages format and apply the appropriate unmasking policy.

Specifically, the new data processing script now reads the `unmask` field to determine whether to unmask all messages (aside from the system message) or just the assistant responses (skills training).

This PR adds:
- A new `process_data` API into the `data_process` module which is meant to be the main backwards-compatible user-friendly entrypoint for data processing
- A new `process_messages_into_input_ids` API in `instructlab.training.data_process` which supports the conversion of any arbitrary HF messages dataset into a masked format that the training library can consume
- Conversion logic to provide compatibility with legacy datasets that contain `pretraining` message roles (based on the format found here: https://github.com/instructlab/sdg/blob/3ca69b6395e8a9cae62f5edaf002f3240a308b66/src/instructlab/sdg/datamixing.py#L380) 
- Updates the training loop to load in a model with the arbitrary tokenizer + chat template
- Example script that shows how someone can call the new data processing APIs

### DIFF
--- a/src/instructlab/training/data_process.py
+++ b/src/instructlab/training/data_process.py
@@ -1,20 +1,32 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
+from enum import StrEnum
 from functools import partial
 from pathlib import Path
 import os
+import typing as t
+import regex as re
 
 # Third Party
-from datasets import load_dataset
-from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
+from datasets import load_dataset, Dataset, disable_caching
+from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast, AutoTokenizer
 import numpy as np
+from tqdm import tqdm
 
 # First Party
-from instructlab.training.config import DataProcessArgs
+from instructlab.training.config import DataProcessArgs, NewDataProcessArgs
 from instructlab.training.tokenizer_utils import get_sp_token, setup_tokenizer
 from instructlab.training.utils import log_rank_0, retrieve_chat_template, setup_logger
 
+# Constants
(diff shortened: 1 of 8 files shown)


src/instructlab/training/data_process.py
```diff
@@ -1,20 +1,32 @@
...
+from instructlab.training.types import Message, ProcessedMessages
 from instructlab.training.utils import log_rank_0, retrieve_chat_template, setup_logger
 
+# Constants
+MASK_TOKEN = "<|MASK|>"
+UNMASK_BEGIN_TOKEN = "<|UNMASK_BEGIN|>"
+UNMASK_END_TOKEN = "<|UNMASK_END|>"
+
+
+disable_caching()
```
Maxusmusti, line 28: NOTE: remove

src/instructlab/training/utils.py
```diff
@@ -201,6 +215,8 @@ def check_flash_attn_enabled(disable_flash_attn: bool, use_dolomite: bool) -> bo
 def make_collate_fn(
     pad_token_id, use_dolomite=False, flash_enabled=True, max_batch_len=60000
 ):
+    # some tokenizers may not have a pad token id, so we use -100 as a default
+    pad_token_id = pad_token_id if pad_token_id is not None else -100
```
Maxusmusti, line 219: NOTE: Remove old artifact

Maxusmusti (approved):
LGTM

aldopareja (approved):

PR #430, opened 2025-03-21
adds Granite Liger Kernel option for Granite3.y models

JamesKunstle:
Support for Granite was added in LK v0.5.4, so we can add it as an additional performance option.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -670,20 +670,17 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         )
 
     if train_args.process_data:
-        # XXX(osilkin): make a decision here, either:
-        #   1. the CLI is fully responsible for managing where the data is written
-        #   2. we never cache it and simply write it to a tmp file every time.
-        #
-        # An important reason for why #1 would be preferable is in the case of OpenShift/SELinux
-        # where the user has a defined place for new temporary data to be written.
+        # TODO(osilkin):
+        #   Decouple the data processing logic from training.
+        #   Now that we've decided that repos will be less tethered to the
+        #   design choices of the `ilab` CLI, we can make this change.
         dp.process_data(
             data_output_path=train_args.data_output_dir,
             model_path=train_args.model_path,
             data_path=train_args.data_path,
             max_seq_len=train_args.max_seq_len,
             chat_tmpl_path=train_args.chat_tmpl_path,
-            num_cpu_procs=torch_args.nproc_per_node,
-            use_legacy_method=train_args.use_legacy_data_processor,
+            num_cpu_procs=train_args.data_process_num_cpu_procs,
         )
 
     if not os.path.exists(train_args.ckpt_output_dir):
--- a/examples/test-dataprocess.py
+++ b/examples/test-dataprocess.py
@@ -1,13 +1,15 @@
-# Standard
-import argparse
-
-# First Party
-from instructlab.training.data_process import process_data
+# SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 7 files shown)


JamesKunstle:
Current CI failure seems unrelated to liger_kernel addition

RobotSail:
Need to test but this could be very useful if it works correctly

src/instructlab/training/main_ds.py
```diff
@@ -35,6 +35,18 @@
     if __name__ == "__main__" and (not local_rank or local_rank == 0):
         print("DeepSpeed is not available. Some features may be unavailable.")
 
+try:
+    # Third Party
+    from liger_kernel.transformers import apply_liger_kernel_to_granite
+except ImportError:
+    apply_liger_kernel_to_granite = lambda *args, **kwargs: None  # pylint: disable=C3001
```
RobotSail, line 42: @JamesKunstle We'll want to support other models beyond Granite, and it seems like Liger's API has a number of different functions they expose for common architectures (mistral, llama, etc.). 

It seems like they provide a way to map directly from the `model_type` field on the model into the Liger kernel to apply. Could we please use that here to support more models? 

https://github.com/linkedin/Liger-Kernel/blob/293bf7eec7043c8c34b3cd82975c97e4c2f4254f/src/liger_kernel/transformers/monkey_patch.py#L1058C1-L1058C29

JamesKunstle:
Currently correctness test for the kernels themselves are done in-tree in the liger-kernels repo. They assure correct convergence and logit equivalence after training. [Link to test here](https://github.com/linkedin/Liger-Kernel/blob/293bf7eec7043c8c34b3cd82975c97e4c2f4254f/test/convergence/bf16/test_mini_models_with_logits.py#L472)

I've also validated training-dynamics equivalence given identical batches. [Link to Jira issue here](https://issues.redhat.com/browse/RHELAI-3102).

There's a roughly 1% raw improvement with batches being identical This is expected- the real benefits will come from larger possible batch sizes! This kernel set is also missing the most important kernel: `LigerFusedLinearCrossEntropy` which skips logit materialization, saving a lot of net memory headroom. We'll add this in the future.

src/instructlab/training/main_ds.py
```diff
@@ -974,6 +995,9 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
             "The last checkpoint will be saved as 'last_epoch'."
         ),
     )
+
+    # this will work for granite-3.y models but not granite-7b because that's a Llama 2 model arch.
+    parser.add_argument("--enable-granite-liger-kernel", action="store_true")
```
RobotSail, line 1000: If we're providing this as an arg through the CLI, we should also expose this in the `TrainingArgs` config so the SDK can be consistent.

RobotSail (commented):
@JamesKunstle Thanks for adding this, a few comments

src/instructlab/training/main_ds.py
JamesKunstle: ++ agreed, will do once we've got the Granite numbers from PSAP

src/instructlab/training/main_ds.py
JamesKunstle: ++ certainly, only limiting to Granite to get PSAP numbers then will expand the integration to other supported architectures.

RobotSail:
Test

RobotSail:
Looks like GitHub forces you to make an initial comment before the option to reopen becomes available
https://github.com/instructlab/training/pull/430#issuecomment-2780915617

RobotSail:
Never mind, it looks like the reopen and comment button is a lie

PR #432, opened 2025-03-24
Generic notebook support

Maxusmusti:
Fixes logging to work in notebook environments, and also removes the existing file path barrier for huggingface models, as a follow-up to @abhi1092's PR #416 

Also now accepts any CausalLM, rather than the previous hard-coded list.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -182,7 +182,9 @@ def setup_model(args, tokenizer, train_loader, grad_accum, flash_enabled):
         )
         model.config.eos_token_id = tokenizer.eos_token_id
 
-    assert "ForCausalLM" in model.__class__.__name__, f"Model class name: {model.__class__.__name__} is not supported."
+    assert (
+        "ForCausalLM" in model.__class__.__name__
+    ), f"Model class name: {model.__class__.__name__} is not supported."
 
     model = convert_loss_to_reduce_sum(model, use_dolomite=args.use_dolomite)
     model = add_noisy_embeddings(model, noise_alpha=args.NEFTune_alpha)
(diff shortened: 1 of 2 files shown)


src/instructlab/training/utils.py
```diff
@@ -65,7 +65,7 @@ def check_valid_train_args(train_args: TrainingArgs):
             raise FileNotFoundError(
                 "Model path does not appear to be a directory. Please make sure that you're passing a Hugging Face Transformers compatible directory checkpoint."
             )
-    else:
+    elif not len(train_args.model_path.split("/")) == 2:
```
RobotSail, line 68: What if I pass `gpt-2`?

src/instructlab/training/main_ds.py
```diff
@@ -182,15 +182,9 @@ def setup_model(args, tokenizer, train_loader, grad_accum, flash_enabled):
...
-        "MistralForCausalLM",
-        "GPTDolomiteForCausalLM",
-        "LlamaForCausalLM",
-        "Starcoder2ForCausalLM",
-        "GemmaForCausalLM",
-        "MixtralForCausalLM",
-        "GraniteForCausalLM",
-    ], f"Model class name: {model.__class__.__name__} is not supported."
+    assert (
+        "ForCausalLM" in model.__class__.__name__
```
RobotSail, line 186: I think we had a reason for this restriction, but I think it will be fine to lift this and revisit if it ever reappears.

src/instructlab/training/utils.py
RobotSail, line 68: What if I pass gpt-2?

RobotSail (approved):
One question but otherwise LGTM

src/instructlab/training/utils.py
Maxusmusti: Updated error message to make expected input more clear

aldopareja (approved):

Issue #437, opened 2025-04-01
Enable support for adding chat templates to base models


RobotSail:
Today in order to train with an arbitrary model, its tokenizer must already have a chat template and an `eos_token` set. 

But in if you want to train a base model with a chat template, you must use our hard-coded chat templates.

We want to implement the ability to give a base model an arbitrary chat template through the following methods: 
- specifying an existing instruct model to use as the target 
- providing some minimal metadata for the model to use (chat template + eos_token + pad token) 

This will allow us to fully deprecate the hard-coded chat template system.

PR #438, opened 2025-04-01
feat: split process_data out from run_training

### LINKED ISSUE #434: refactor data processing to enable passing output file path
Today the data processing logic will write to `"{output_path}/data.jsonl"`, but we want to adjust this so that users can simply pass a file path that is assumed to be an output filepath.

cdoern:
process_data=True was allowing users to implicitly process their data via run_training since `ilab` isn't the main consumption point of the training library going forward, it makes sense to separate these concerns.

remove `data_output_dir` and assume `data_path` is processed data. also remove the `process_data` argument.

Adjust documentation, notebooks, etc

resolves #434, #437


RobotSail (approved):
LGTM

README.md
```diff
@@ -335,13 +334,14 @@ from instructlab.training import (
     DataProcessArgs,
     data_process as dp
 )
+import os
```
booxter, line 337: nit, Opinion, Not required: in some codebases I worked with, it's common to import `os.path` instead of `os` to give a hint that we care about the path functions specifically and not just any `os` stuff.

src/instructlab/training/config.py
```diff
@@ -162,13 +162,10 @@ class TrainingArgs(BaseModel):
...
     use_legacy_tmpl: bool = False
 
-    # this field specifies the filepath to the training dataset before processing
+    # this field specifies the filepath to the training dataset
     data_path: str
-    ckpt_output_dir: str
 
-    # this field defines where we should be saving the processed version of the training dataset
-    # after we have tokenized it
-    data_output_dir: str
```
booxter, line 171: AFAIU instructlab CLI currently uses this field to pass --data-output-dir from the user. It also assumes the default behavior - `process_data` being `True`.

Would it be possible / desirable to make the transition smoother, e.g. by

1. marking `data_output_dir` field as deprecated;
2. if it's passed, direct users with a warning to use the data_path name instead (plus not assume pre-processing?).

Or is the strategy here is to rewrite CLI + bump the minimal version for training library? Even if the CLI training rewrite for the new interface is the plan, it would be easier on the remaining CLI maintainers if the change happens through gradual deprecation.

booxter:
You mentioned #437 in the PR description as resolved. I am not sure I understand the relation between them. Would you mind clarifying how they are related for those new to the repo? Thanks.

src/instructlab/training/config.py
RobotSail: > Even if the CLI training rewrite for the new interface is the plan, it would be easier on the remaining CLI maintainers if the change happens through gradual deprecation 

I'm not sure how this is handled in the current refactor @cdoern  is working on. Assuming it doesn't handle this case, this is how I recommend we handle it based on how we made the initial refactor to move to the newer version of data processing:

1. Create a new API for what is currently `run_training` which only runs training, we can either call it `run_training_v2` or something else. 
2. Keep `run_training` with a deprecation notice and have it still perform this orchestration under the hood
3. Refactor `ilab` to use the new APIs
4. Remove `run_training` after a few releases as you have described here

README.md
RobotSail: @booxter or better yet, `from os import path` 😉

PR #443, opened 2025-04-07
fix: model.forward now accepts return_dict via kwargs

booxter:
Since https://github.com/huggingface/transformers/pull/36794 it handles
the return_dict with can_return_tuple decorator that expects the
argument to be passed as kwarg.

Without the patch, instructlab functional tests fail with:

AttributeError: 'bool' object has no attribute 'unsqueeze'

This is because the passed return_dict argument is incorrectly
interpreted as cache_position.

Of course, the way the model forward method is overridden here is quite
problematic and would benefit from a refactor. This patch is a stop-gap
to fix the CI with minimal changes.

### DIFF
--- a/src/instructlab/training/utils.py
+++ b/src/instructlab/training/utils.py
@@ -396,16 +396,16 @@ def convert_loss_to_reduce_sum(model, use_dolomite=False):
             **deprecated_arguments,
         ):
             output = model.__original_forward__(
-                input_ids,
-                attention_mask,
-                position_ids,
-                past_key_values,
-                inputs_embeds,
-                labels,
-                use_cache,
-                output_attentions,
-                output_hidden_states,
-                return_dict,
+                input_ids=input_ids,
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                past_key_values=past_key_values,
+                inputs_embeds=inputs_embeds,
+                labels=labels,
+                use_cache=use_cache,
+                output_attentions=output_attentions,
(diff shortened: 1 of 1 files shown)


booxter:
@JamesKunstle please take a look. I am not sure about the change since I don't grasp the full extent of the override; I think I understand the root cause and I did my best to clean the code up from mere mechanical code reading. Someone with proper expertise would need to check if this makes in a broader context of what the code is trying to achieve beyond... not crashing.

booxter:
Apparently e2e failure was supposed to be fixed by https://github.com/instructlab/instructlab/pull/3254 but since this PR fixes functional tests for that other PR, we have chicken and egg. I guess instructlab will have to pin / cap transformers library first; then we can later land this change here; then finally bump in instructlab and uncap transformers.

JamesKunstle (approved):
lgtm!

src/instructlab/training/utils.py
```diff
@@ -396,16 +396,16 @@ def reduce_sum_forward(
...
+                past_key_values=past_key_values,
+                inputs_embeds=inputs_embeds,
+                labels=labels,
+                use_cache=use_cache,
+                output_attentions=output_attentions,
+                output_hidden_states=output_hidden_states,
+                return_dict=return_dict,
             )
 
             return_dict = isinstance(output, dict)
```
booxter, line 411: This actually makes no sense since it's never a dict but tuple, maybe.

RobotSail (approved):

PR #455, opened 2025-04-14
moves deepspeed requirements into their own file; add deepspeed extras

JamesKunstle:
users may not want to install deepspeed when they install the cuda requirements. this moves DeepSpeed into its own requirements file so if the user wants to install DeepSpeed, they can do so with:

`pip3 install instructlab-training[cuda,deepspeed]`

### DIFF
--- a/requirements-deepspeed.txt
+++ b/requirements-deepspeed.txt
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+
+deepspeed>=0.14.3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = { file = ["requirements.txt"] }
 optional-dependencies.cuda = { file = ["requirements-cuda.txt"] }
 optional-dependencies.rocm = { file = ["requirements-rocm.txt"] }
 optional-dependencies.hpu = { file = ["requirements-hpu.txt"] }
+optional-dependencies.deepspeed = { file = ["requirements-deepspeed.txt"] }
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/requirements-cuda.txt
+++ b/requirements-cuda.txt
@@ -1,9 +1,6 @@
 flash-attn>=2.4.0
 bitsandbytes>=0.43.1
 
-# available as an option for NVIDIA, FSDP still default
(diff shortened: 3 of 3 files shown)


JamesKunstle:
Needed for version request: https://issues.redhat.com/browse/RHELAI-3782

requirements-deepspeed.txt
```diff
@@ -0,0 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+
```
RobotSail, line 2: If we are installing deepspeed then this should automatically also install the CUDA target no?

JamesKunstle: The build step doesn't allow for chaining dependencies in the files, weirdly.

e.g. you can't have `requirements-deepspeed.txt` be 
```
-r requirements-cuda.txt
deepspeed>=0.14.3
```
The build system complains. 

Instead, you have to install the extras together with
`pip3 install instructlab-training[cuda,deepspeed]` or `pip3 install ".[cuda,deepspeed]"`

I'm looking into whether we can group `-cuda.txt` and `-deepspeed.txt` in the tox.ini so that they get installed together if someone installs deepspeed.

booxter:
FYI the flash-attn issue in CI doesn't seem related to this PR after all, I hit it elsewhere: https://github.com/instructlab/training/actions/runs/14480518080/job/40616430409?pr=467

JamesKunstle:
At least philosophically, I guess the extras installation should be:
`pip3 install instructlab-training[cuda,deepspeed]`
or 
`pip3 install instructlab-training[rocm,deepspeed]`

If `deepspeed` always pulls the requirements for `cuda` that doesn't make sense.

RobotSail (approved):
Now that #465 has merged, we can make this change

tiran:
> At least philosophically, I guess the extras installation should be: `pip3 install instructlab-training[cuda,deepspeed]` or `pip3 install instructlab-training[rocm,deepspeed]`
> 
> If `deepspeed` always pulls the requirements for `cuda` that doesn't make sense.

+1

James' approach makes sense and is an established pattern.

cdoern:
@mergifyio backport release-v0.10

ktdreyer:
Do we need to backport this to `release-v0.11` too?

PR #456, opened 2025-04-15
feat: refactor main_ds.py

cdoern:
into classes:

1. Model
2. Accelerator
4. Checkpointer

**NOTE: a follow up to this work will be to introduce classes/structure for the DataLoader, Sampler, etc. This was left out of this PR given the already large scope of change.** 

The **Model** class wraps the various AutoModel classes we support -- and aims to be a lightweight wrapper to help with usability of the library with different model types.
setup_optimizer resides within the model class and returns one of the optimizer types we support

The **Accelerator** class aims to both store commonly accessed variables associated with the accelerated model and abstract model/optimizer mutation away from the user who should only access our Model and Optimizer classes

The **Checkpointer** class introduces a unified approach to our various checkpointing techniques. A user can pass in their checkpointing style (full_state or hf_format), and the checkpointer, via checkpointer.checkpoint, will save the model using the selected method and other techniques (LoRA).

Being able to group functionality under the Model, Optimizer, Accelerator, and Checkpointer classes inherently reduces code complexity and duplication. Being able to store things like , `self.distributed_framework`,`self.lora_config`, etc in a way such that within the class they are accessible within different methods allows the arguments per method to go down drastically, as well as complex return values. Simpler methods and argument/return values allows for simpler testing of code.


fynnsu:
Looks good! A couple high level suggestions:
- I like the use of model, accelerator, and checkpointer classes but I am not sold on the optimizer class right now. In its current form it's essentially a passthrough wrapper class. I think because there is slightly better standardization for optimizers in PyTorch vs models/accelerators/checkpointers the optimizer classes apis are fairly consistent already and therefore there may be less need for this class. I think instead the optimizer init method could just be a function that returns either AwamW/DeepSpeedCPUAdam/FusedAdam directly.
- Some of the function/method argument lists are pretty hefty (`Model.__init__`/ Model class methods, the functions in `train.py`, etc). For the `train.py` functions, I get that this is partially a consequence of not using a `Trainer` class but I still think these can be cleaned up a bit. I'm personally a big fan of lightweight classes or namedtuples for grouping stuff like this. For example, the Model init/constructor methods could be simplified by initializing the `LoraConfig` in `main_ds.py` and then passing that (or None) to the Model init methods, instead of 4 Lora specific arguments. I know the `create_lora_config` method also has additional logic for limiting the target modules to valid ones, but this logic could be moved into the `prepare_peft_model` util fn instead. And similarly maybe some kind of `train_config` object can be used to group a lot of the fixed arguments in `train.py` together.

cdoern:
@fynnsu thinking through further refactoring using `LoraConfig` as an example:

Currently `Model` has a `create_lora_config` method which associates a lora_config with a `Model` instance.

if we were to move this out and force users to create their own `LoraConfig` which forces the user to create their own `AutoModel` which is pretty confusing if a user is creating an `AutoModel` and a `Model` class  is there much gain over passing these few extra variables?

I can go through and see if there are other examples, but the reason I structured this class the way I did is because `self.model` is needed for a _bunch_ of initialization steps so if we split things out of the __init__ we risk re-introducing the "ravioli" code where we pass different "models" around and now it'll be worse since one is of type `Model` (ours) and one will be an `AutoModel`.

cdoern:
closing in favor of #498

PR #459, opened 2025-04-15
chore: Run tests when running `make tests`

booxter:


JamesKunstle:
This change is good, although I'm also in favor of just removing the rule and using tox directly.
I guess someone might want `make tests` to just run `tox -e py3-unit,py3-smoke` but that seems like an arbitrary distinction when `tox` is so much simpler.

JamesKunstle (approved):
I agree and would merge after comment gets feedback

booxter:
Updated PR to just remove the target completely.

RobotSail (changes requested):
Why are we removing this? I use this command all of the time

RobotSail (changes requested):
Thank you for updating this @booxter . Can we also please create either a `CONTRIBUTING.md` or update the `README.md` so new contributors are aware how they can run tests in this repo? Otherwise we will just have this and nobody will know how to use it.

booxter:
IMHO if `CONTRIBUTING.md` document should be added, it's out of scope for this PR. I probably won't have time to write the document, so if it's a requirement here, feel free to close the PR.

RobotSail:
> IMHO if CONTRIBUTING.md document should be added, it's out of scope for this PR. I probably won't have time to write the document, so if it's a requirement here, feel free to close the PR.

I'm not really following what you mean. It shouldn't take more than 5 minutes to put together a simple document that says something like 


## Contributing

If you are developing on this repo, you can test your code by running these commands:
```
pip install -r requirements-dev.txt
make tests
```

booxter:
OK if that's all you are looking for. I thought you want to include something more elaborate. Done.

CONTRIBUTING.md
```diff
@@ -0,0 +1,7 @@
+## Contributing
+
+If you are developing on this repo, you can test your code by running these commands:
```
RobotSail, line 3: @booxter Could you please expand this section to include a few more thing? This general format is fine, but we should include the few other tools we have so that it's easier for new contributors to get started:

1. Users can generally run `tox` or `make tests` to just run everything
2. Before making a commit, users can check their code for correctness with `tox -e lint` and `tox -e mypy`
3. All trivial errors can be fixed with `tox -e ruff`

PR #463, opened 2025-04-15
chore: don't trigger unit tests for cuda and rocm requirements changes

booxter:

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -18,7 +18,8 @@ on:
       - "**.py"
       - "tox.ini"
       - "pyproject.toml"
-      - "requirements**.txt"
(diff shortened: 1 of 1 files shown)


RobotSail (commented):
I think I'm missing some context here but why would we want to do this? What if there's a package being installed through one of those libraries for some reason and the change creates bugs elsewhere?

booxter:
@RobotSail only requirements.txt and -dev.txt are used for unit tests: https://github.com/instructlab/training/blob/2519deaf4e1a69925d3d42b43e62f7fb4994dab4/tox.ini#L25C8-L25C28

RobotSail (approved):
LGTM

JamesKunstle (approved):

PR #465, opened 2025-04-15
Fix FSDP mixed precision setting and loss w/ accelerate

Maxusmusti:
Accelerate requires mixed_precision to be passed directly in config, else the fp32 upcast gets skipped and training loses overall precision, resulting in notably worse performance. Passing this also makes our manual mixed precision policy redundant, and since we are now _actually_ using mixed precision, hybrid shard finally becomes the new default sharding strategy to keep memory expectations in parity.


RobotSail (approved):

JamesKunstle (approved):
👍

PR #467, opened 2025-04-15
chore: Drop pytest-asyncio

booxter:


JamesKunstle:
Why do we need the asyncio stuff in this lib anyway? I've been wondering if we can just drop that requirement altogether

booxter:
We can definitely drop it at least until we feel the need. I will drop.

RobotSail:
@JamesKunstle We use async logic in order to optimize I/O logic and optimize the amount of CPU time available so the training loop doesn't hang on blocking FS write calls. You can see this in action if you look at the AsyncStructuredLogger implementation here: https://github.com/instructlab/training/blob/main/src/instructlab/training/async_logger.py

RobotSail (approved):
We can remove this now, but I have a feeling we will need to just bring it back in again later.

JamesKunstle (approved):
We can re-add this when we have tests that need it. For now it's tidier to just remove it.

PR #470, opened 2025-04-15
chore: Require python3.11+

booxter:
We are not testing on older versions, and there's not a lot of reasons
to do so anyway. I believe for RHEL AI, we care about Python 3.11+ only.

For reference, https://devguide.python.org/versions/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ authors = [{ name = "InstructLab", email = "dev@instructlab.ai" }]
 description = "Training Library"
 readme = "README.md"
 license = { text = "Apache-2.0" }
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 classifiers = [
   "Development Status :: 3 - Alpha",
   "Environment :: Console",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [{ name = "InstructLab", email = "dev@instructlab.ai" }]
 description = "Training Library"
 readme = "README.md"
 license = { text = "Apache-2.0" }
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
RobotSail, line 13: We only officially support Python3.11 today. Until that changes, this should be capped.

RobotSail (changes requested):

pyproject.toml
booxter: This is not the goal of this patch (which is to drop support for 3.10). But also upper capping python version is considered harmful. It's quite a long story to go in the comments here, so I'll instead share this [nice blog](https://iscinumpy.dev/post/bound-version-constraints/) that covers capping intricacies. (The exact section covering why requires-python should not be capped can be found under "Pinning the Python version is special" title.)

RobotSail: Thanks for the blog @booxter , that makes a lot of sense. This is fine then.

RobotSail (approved):
LGTM

JamesKunstle (approved):

PR #471, opened 2025-04-15
Enable miscellaneous ruff rules

booxter:
- **chore: Enable flake8-bugbear checks**
- **chore: Enable Pyflakes checks**
- **chore: Enable pyupgrade**
- **chore: Enable flake8-simplify**

### DIFF
--- a/src/instructlab/training/utils.py
+++ b/src/instructlab/training/utils.py
@@ -278,10 +278,13 @@ def make_collate_fn(
                 total_len = 0
                 num_loss_counted_tokens = 0
 
-                for num_samples, item in enumerate(batch):
+                num_samples = 0
+                for item in batch:
                     item_len = len(item["input_ids"])
                     if total_len + item_len > max_batch_len:
                         break
+
+                    num_samples += 1
 
                     input_ids.extend(item["input_ids"].tolist())
                     labels.extend(item["labels"].tolist())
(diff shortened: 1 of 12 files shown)


RobotSail:
Why do we need this?

src/instructlab/training/utils.py
```diff
@@ -732,7 +723,8 @@ class UniversalCheckpointArgs:
         if UNIVERSAL_CHECKPOINT_INFO not in ds_checkpoint.global_state:
             warnings.warn(
                 "Universal checkpoint information not found, setting it to "
-                "an empty dictionary."
+                "an empty dictionary.",
+                stacklevel=2,
```
booxter, line 727: these should probably be stacklevel=1 after all

booxter:
Reasons:

- modernize code base for newer python features (e.g. not importing Optional);
- fixing some coding issues (e.g. prepare_peft_model was using a mutable dict as default value for an argument, which is unsafe; another code was returning from under `finally` which potentially swallows exceptions).

src/instructlab/training/utils.py
```diff
@@ -99,19 +100,16 @@ def check_valid_train_args(train_args: TrainingArgs):
...
-            import flash_attn
-        except ImportError as exc:
-            raise ImportError(
-                "Flash attention is enabled but flash_attn is not installed. You can resolve this in the following ways:\n"
-                "1. Ensure the CUDA/ROCM version of the training library is installed via: `pip install instructlab-training[cuda]` or `pip install instructlab-training[rocm]`\n"
-                "2. Install flash_attn manually via: `pip install flash-attn --no-build-isolation`\n"
-                "3. Disable flash attention by setting `disable_flash_attn=True` in your training arguments\n"
-            ) from exc
+    # also verify that the flash_attn package is actually installed
+    if check_flash_attn_enabled(
```
booxter, line 104: TODO: confirm this detection works as I expect

PR #473, opened 2025-04-15
ci: Use pull_request as trigger for unit tests

booxter:
Otherwise the tested fork with the PR on review is not validated in CI.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -9,7 +9,7 @@ on:
       - main
       - release-*
   # only run on PRs that touch certain regex paths
-  pull_request_target:
+  pull_request:
     branches:
       - main
       - release-*


booxter:
@courtneypacheco @ktdreyer please confirm this is safe / correct.

ktdreyer:
From reading https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git , I don't think this is safe.

ktdreyer:
I did more reading here: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

Now I have learned more about the differences here:

`pull_request` will run on the PR's code, and it will not have access to any the secrets in our repositories (so it won't be able to run AWS tests).

`pull_request_target` *does* have access to secrets, but we should not use it to run untrusted PR code.

booxter:
@ktdreyer since I switched unit tests to use github runners, is it safe to switch unit tests job to pull_request? Will it give us proper validation for PRs under review?

ktdreyer (approved):
Yes, thanks for clarifying.

I see https://github.com/instructlab/training/pull/475 cleans up the ec2 reference, and I understand this only uses GH's own runners now.

PR #479, opened 2025-04-18
chore: Allow numpy >= 2.0.0

booxter:
A large e2e run with forced numpy 2.0+ suggests that the library is compatible with numpy 2.x series.

Also, removed `numpy` dependency from requirements-dev.txt because it's already listed in requirements.txt.

See: https://github.com/instructlab/training/actions/runs/14539292904

### DIFF
--- a/constraints-dev.txt
+++ b/constraints-dev.txt
@@ -109,7 +109,7 @@ nodeenv==1.9.1            # via pre-commit
 notebook==7.4.3           # via jupyter
 notebook-shim==0.2.4      # via jupyterlab, notebook
 numba==0.61.2             # via -r requirements.txt
-numpy==1.26.4             # via accelerate, bitsandbytes, contourpy, datasets, deepspeed, matplotlib, numba, pandas, peft, tensorboard, transformers, -r requirements-dev.txt, -r requirements.txt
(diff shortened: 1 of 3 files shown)


RobotSail:
@booxter We use numpy in a few critical points like the collator function in the dataloader as well as the data processing logic when converting from messages into raw input IDs for the model to consume. This is a PR where we'd need to run training and validate that the loss curve from your changes produces the same loss curve as what's in the main branch.

booxter:
Fair enough. We'll need to automate validation for patches like this one (also refactors etc.) I don't feel we should continue with human assessments (or at the very least artifacts - charts etc. - needed for such assessment should be generated and posted in comments.) I will mark this as a draft and report an issue to implement the necessary automation.

RobotSail:
@booxter Yes, we have a script to do this already but I think it broke and I haven't had time to look into it. If you can fix it then that would be a huge contribution which would automate most of the testing we do manually today. You can find it here: https://github.com/instructlab/training/blob/main/scripts/create-loss-graph.py 

And then the actual CI code to call it can be found here: https://github.com/instructlab/training/blob/65056fa539a49606aed435533788f2b4370d0e8d/.github/workflows/e2e-nvidia-l40s-x4.yml#L387

booxter:
@mergifyio rebase

RobotSail:
@booxter Sorry I didn't see your message prior as I was out on PTO. Ideally we just need to check what APIs we are using from numpy and whether any are impacted. 95% likely it's fine but would be good to double-check.

PR #482, opened 2025-04-19
ci: Use tox-current-env to reuse prepared venv with torch

booxter:
This is a completion of a workaround for flash-attn. The initial attempt didn't work because we still used tox to run smoke tests, which means another venv was created that did not include flash-attn.

To achieve sharing the venv for both torch installation and smoke tests, we are using tox-current-env plugin here.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -143,10 +143,23 @@ jobs:
       # setup.py depends on importing the module, so it should have been listed
       # in build_requires. Alas.
       # See: https://github.com/Dao-AILab/flash-attention/pull/958
-      - name: "Install torch before other dependencies"
+      - name: "Install torch and other unlisted build dependencies for flash-attn"
         run: |
           source venv/bin/activate
-          pip install torch
+          # The list is taken from the pull request linked above
+          pip install torch packaging setuptools wheel psutil ninja
+
+      - name: "Install tox-current-env to reuse the venv with pre-installed build dependencies"
+        run: |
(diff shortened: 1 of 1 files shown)


RobotSail (commented):
@booxter Is the problem that this fixes only in CI? Or would the end user be expected to run tox in the same way (i.e. having a `venv`)?

booxter:
@RobotSail until https://github.com/Dao-AILab/flash-attention/pull/958 is merged (and they don't seem keen to do so for reasons that I don't agree with), every project that depends on flash-attn has to pre-install their build dependencies, particularly pytorch. `tox` doesn't allow to order installation of dependencies to guarantee that torch is always installed before flash-attn, so I think any `tox` user is potentially affected by this. `tox` being a development tool and not something an end-user is supposed to use in production is probably a mitigating factor, but yes - this sucks. :(

booxter:
Testing the change here: https://github.com/instructlab/training/actions/runs/14713551885

booxter:
(The more I patch this, the more I don't like it. But let's see how horrible the final "fix" would look like...)

booxter:
This run confirms that tox-current-env works now: https://github.com/instructlab/training/actions/runs/14719818289/job/41311408922

Now to decide if we want to see this solution. (If not, I'm all ears to hear about alternatives that would allow us to run smoke tests with `[cuda]` depenendencies that include the `flash-attn` library with broken build instructions.)

RobotSail (approved):

nathan-weinberg (approved):

PR #483, opened 2025-04-20
use correct field name for mock_data_len

RobotSail:
Changes the `run_training` script to use the correct field name when referencing `mock_data_len`.

Signed-off-by: eshwarprasadS <eshwarprasad.s01@gmail.com>


src/instructlab/training/config.py
```diff
@@ -228,3 +229,9 @@ class TrainingArgs(BaseModel):
         default=False,
         description="Whether to use Liger kernels for training.",
     )
+
+    # TODO(osilkin): Create a better API for this, should not merge into library this way
```
cdoern, line 233: I know Fynn has been working on "SDK-ifying" the sampler specifically @RobotSail , maybe we should sync on this with the training team

PR #486, opened 2025-04-21
Obey the FSDP sharding option default

Maxusmusti:
Our default FSDPOptions object in training config was overriding the default we set in the FSDPOptions object definition.


RobotSail (approved):

Maxusmusti:
@Mergifyio backport release-v0.10

Issue #494, opened 2025-04-24
e2e large ci failures


ktdreyer:
This CI workflow is currently failing on `main`: https://github.com/instructlab/training/actions/workflows/e2e-nvidia-l40s-x4.yml

Last passing run was five days ago: https://github.com/instructlab/training/actions/runs/14550704806

@courtneypacheco speculated that a dependency may have changed recently that broke this CI workflow.

Next step: compare the dependencies we installed in the green run with the dependencies we installed in the first failed run.

ktdreyer:
After our last success (5 days ago):

The very [first failure (5 days ago)](https://github.com/instructlab/training/actions/runs/14561125588/job/40844613726) has this line:
```
##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 5 MB
```

Our most [recent failure (an hour ago)](https://github.com/instructlab/training/actions/runs/14649266088/job/41111106399) does not have this disk space error any more. It does have this:

```
RuntimeError: Failed to import transformers.models.rt_detr.modeling_rt_detr because of the following error (look up to see its traceback):
/actions-runner/_work/training/training/instructlab/venv/lib64/python3.11/site-packages/flash_attn_2_cuda.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZN3c105ErrorC2ENS_14SourceLocationENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```

ktdreyer:
Comparing the [successful `e2e-large-test` job (5 days ago)](https://github.com/instructlab/training/actions/runs/14550704806/job/40821584662): 
With the [failed `e2e-large-test` job (an hour ago)](https://github.com/instructlab/training/actions/runs/14649266088/job/41111106399):
```
+---------------------------+-----------------+----------------+
|          Package          | Success Version |  Fail Version  |
+---------------------------+-----------------+----------------+
|          aiohttp          |     3.11.16     |    3.11.18     |
|           boto3           |     1.37.37     |     1.38.1     |
|          botocore         |     1.37.37     |     1.38.1     |
|        docling-core       |      2.27.0     |     2.28.0     |
|     docling-ibm-models    |      3.4.1      |     3.4.2      |
|          greenlet         |      3.2.0      |     3.2.1      |
|        haystack-ai        |      2.12.2     |     2.13.1     |
|           hf-xet          |      1.0.3      |     1.0.4      |
[...]

PR #495, opened 2025-04-24
fix(deps): Re-install flash-attn with '--no-build-isolation' to temporarily unblock CI

courtneypacheco:
There appears to be an error related to `flash-attn`. More info here: https://github.com/instructlab/training/issues/494

This PR is a temporary solution to unblock the large E2E job.


ktdreyer:
I think we should take this approach instead: https://github.com/instructlab/training/pull/499

booxter:
Should be fixed with https://github.com/instructlab/training/pull/499 Feel free to reopen if the issue is still present.

PR #497, opened 2025-04-25
chore: introduce dummy workflow

cdoern:
a workflow must be registered before it can be work on in a PR.

Introduce dummy file to be worked on in `refactor` branch


booxter:
This broke CI...

PR #498, opened 2025-04-25
feat: refactor main_ds.py

cdoern:
into classes:

1. Model
2. Accelerator
3. Checkpointer

NOTE: a follow up to this work will be to introduce classes/structure for the DataLoader, Sampler, etc. This was left out of this PR given the already large scope of change.

The **Model** class wraps the various AutoModel classes we support -- and aims to be a lightweight wrapper to help with usability of the library with different model types.
setup_optimizer resides within the model class and returns one of the optimizer types we support

The **Accelerator** class aims to both store commonly accessed variables associated with the accelerated model and abstract model/optimizer mutation away from the user who should only access our Model and Optimizer classes

The **Checkpointer** class introduces a unified approach to our various checkpointing techniques. A user can pass in their checkpointing style (full_state or hf_format), and the checkpointer, via checkpointer.checkpoint, will save the model using the selected method and other techniques (LoRA).

Being able to group functionality under the Model, Optimizer, Accelerator, and Checkpointer classes inherently reduces code complexity and duplication. Being able to store things like , self.distributed_framework,self.lora_config, etc in a way such that within the class they are accessible within different methods allows the arguments per method to go down drastically, as well as complex return values. Simpler methods and argument/return values allows for simpler testing of code.


RobotSail:
@cdoern I haven't gone through this in-depth yet, but I'm wondering if it would make sense to build into a v1 and do a major version bump to `v1.0.0`.

I'm not sure how much you are looking to break APIs that we have today, but if we went down this path then it would be a very good time to explore this option. This way we can work it into something nicer and ergonomic compared to how it is today.

Since I think it would also be unfair for you to be the only person working on this, and have the PR sitting while changes are actively being made in the repo, I propose this:

1. We create a new branch `training-v1` where we can merge this in as soon as you're ready. This way myself and anyone else can start using it and helping to build on top of it
2. We can begin cutting dev releases such as `instructlab-training-v1.0.0a1`, `instructlab-training-v1.0.0.preview`, etc. to consume this and use it
3. We should identify any other bottlenecks of the current library and address them in this branch (only big things that are difficult to change later, we shouldn't hold this for too long)
4. We can create compatibility APIs so that any consumers (primarily `ilab`) aren't affected, similar to what we've done for the data processing refactor and what @booxter was suggesting on the other PR
5. Once we are ready, we can merge into main and cut a new release

cdoern:
@RobotSail sail I think that is a great idea! Thank you for looking at this, I really wanted to get the loss gathering working to prove at the very least I wasn't degrading anything -- but a branch other than main makes complete sense to me so we can begin iterating on this!! I can open a different PR to that branch as opposed to the main branch here. Again, thanks for thinking about this and taking a look at the PR :)

src/instructlab/training/model.py
```diff
@@ -0,0 +1,1077 @@
...
+        log_rank_0(f"\033[93mModel saved in {final_output_dir}\033[0m", to_print=True)
+        log_rank_0(f"saving took {time.time() - start} seconds")
+        dist.barrier()
+
+        self.accelerator.get_state_dict = get_state_dict_unpatched
+
+    def save_dict_accelerate(
+        self,
+        accelerator: Accelerator,
+        state_to_save,
```
fynnsu, line 990: It would also be useful to have type hints for these methods. This is potentially not your fault as I imagine you mostly just moved this code around but this method is currently only called with `state_to_save` as a model's `state_dict()`. With that type info, setting `state_to_save.modules` and `state_to_save.parameters` to a noop doesn't make sense because those attributes don't exist in the first place.

src/instructlab/training/model.py
```diff
@@ -0,0 +1,1077 @@
...
+        if self.accelerator.distributed_framework == "fsdp":
+            self.accelerator.get_state_dict = get_state_dict_unpatched
+
+    def save_hf_format_accelerate(
+        self,
+        output_dir,
+        epoch: int,
+        samples_seen: int,
+        last_epoch: bool = False,
+        **kwargs,
```
fynnsu, line 872: `epoch` and `kwargs` not used

src/instructlab/training/model.py
```diff
@@ -0,0 +1,1077 @@
...
+                self.model.base_model_args["device_map"] = old_device_map
+
+        dist.barrier()
+
+    def save_full_state(
+        self,
+        output_dir,
+        epoch: int,
+        samples_seen: int,
+        **kwargs,
```
fynnsu, line 816: `kwargs` not used. I also just raised #527 to add checks for unused parameters to our linter.

src/instructlab/training/model.py
```diff
@@ -0,0 +1,1077 @@
...
+        return cls(
+            model=model,
+            grad_accum=grad_accum,
+            train_loader=train_loader,
+            distributed_framework=DistributedBackend.FSDP,
+            samples_per_gpu=samples_per_gpu,
+            fsdp_sharding_strategy=fsdp_sharding_strategy,
+            fsdp_cpu_offload_params=fsdp_cpu_offload_params,
+            save_samples=save_samples,
+        )
```
fynnsu, line 718: Maybe these two classmethods should just use `**kwargs`?

src/instructlab/training/model.py
```diff
@@ -0,0 +1,1077 @@
...
+        # Local
+        from .utils import add_noisy_embeddings, convert_loss_to_reduce_sum
+
+        self.model = convert_loss_to_reduce_sum(
+            self.model, use_dolomite=(self.model_type == "dolomite")
+        )
+        self.model = add_noisy_embeddings(self.model, noise_alpha=self.noise_alpha)
+
+    @staticmethod
+    def supports_flash_attention(device_id=0):
```
fynnsu, line 406: I think this staticmethod and the next `check_flash_attn_enabled` were better as functions in `utils.py`. They're only loosely model related, really they're system related. And they don't use any attributes from `Model`

fynnsu (commented):
Looks good. Added a few suggestions.

src/instructlab/training/train.py
```diff
@@ -0,0 +1,238 @@
...
+    if accelerator.save_samples > 0:
+        accelerator.save_samples = (accelerator.save_samples // batch_size) * batch_size
+        (
+            print(
+                f"\033[93mNumber of samples per save: {accelerator.save_samples}\033[0m"
+            )
+            if local_rank == 0
+            else None
+        )
+    all_metrics: List[Metrics] = []
```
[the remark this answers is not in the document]
RobotSail: Yeah it in general seems there is a lot of overlapping work that you've done @fynnsu which is better suited to be built on top of the new version @cdoern is building out here, so I wonder if it would be easier to merge this into a dev branch, merge the other PRs on there, then release it as main once everything is ready.

booxter (changes requested):
I've looked at the patch from birds view. I haven't checked the logic in details. This is because I believe this PR will have to be split into separately reviewable changes before it can be considered for inclusion. Probably a series of commits, each doing one logical transformation.

I won't give you the immediate recipe on the exact split that would make sense here - after all, you probably know better which logical steps this large PR consists of than anyone else - but as a generic example, here's what I would expect from a PR of this size and impact (or a set of such PRs):

* a commit moving a unit of code from module A to B (with no changes to what the code does);
* a commit cleaning up the unit of code (e.g. cleaning up spurious logging);
* perhaps another commit cleaning up some other issue in the moved code;
* another commit introducing a stub class Model that demonstrates the interface but has no implementation, with some basic unit tests creating an object of the type;
* another commit plugging the previously moved and transformed unit of code into one of the methods of the new class, with additional unit test for this method, ...;
* ..._more commits in between of the similar nature_...
* a separate commit introducing a tox target triggering an integration test (the e2e thingy included here);
* a separate commit refactoring existing smoke workflow moving the code to initialize a virtual environment with necessary cuda dependencies and ilab into a reusable script;
* a separate commit adding a CI job triggering the script above + new tox target;
* another commit with docs...

Each of the commits would then describe in its message what is being done (can be as simple as "Move X function to model Y").

This is just an outline trying to show how this would look like in general terms. These commits can then be lumped into a single PR or - preferrably? - multiple (so that we can start merging pieces that are ready as they shape up; each PR probably combining a number of these smaller separately reviewable commits).

In this way, we will be able to give proper attention to each of the changes included here. Without it, we would have to take the refactor execution at faith value, which is a risk for regressions.

(I don't think going the route of a feature branch for this work is feasible either. This is because it won't make it any easier to review this work in a single piece **later** when integration from the branch back to main would have to happen. So it will only delay - and exacerbate - the process.)

---

If you need help figuring out how the split should be executed - from the code logic or git perspective - let's discuss this. I have some experience dealing with large refactors (from both sides).

cdoern:
@booxter thanks for the in depth comment! I will think of how to split this up I would imagine it'd be something like:

1. move model initialization code from main_ds.py to a new module, add unit tests
2. move optimizer initialization code to model module, add unit tests
3. move accelerator code from main_ds.py to new module, add unit tests
4. create checkpointer class concept and move associated code from main_ds.py and utils.py into model module
5. add new e2e test which runs SDK.


I can work next week on splitting up those changes into separate PRs

JamesKunstle:
This PR is being broken down into smaller modular PRs by @cdoern

PR #499, opened 2025-04-25
use instructlab `constraints-dev.txt` in e2e test

ktdreyer:
For instructlab, `pip install .` does not install `vllm`, but it does install an uncapped `torch` (`2.7.0` currently).

When we install `vllm` later, we compile a binary `flash_attn` wheel against `torch-2.7.0`. `vllm-0.8.4` requires `torch==2.6.0`, so we downgrade `torch`, and then we use that with the incompatible `flash_attn` binary wheel.

To resolve this, use instructlab's `constraints-dev.txt` in the first `pip install` operation. This restricts `torch` to `2.6.0` immediately when we first install instructlab, so that we will compile `flash_attn` against that `torch` version.

Fixes: #494

### DIFF
--- a/.github/workflows/e2e-nvidia-l40s-x4.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x4.yml
@@ -185,7 +185,7 @@ jobs:
           nvidia-smi
           python3.11 -m pip cache remove llama_cpp_python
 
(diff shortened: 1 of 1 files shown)


booxter (approved):
Correct fix. Smoke tests are affected by something else (even if adjacent.)

JamesKunstle (approved):

PR #500, opened 2025-04-25
Add general logging implementation

fynnsu:
Adds a general metric logger format with support for logging to json file, wandb, tensorboard, and the existing async logger. 

User can select the `--logger_type`:
- `async`: uses the existing AsyncStructuredLogger, this is the default option and default file names are the same for backwards compatibility reasons
- `file`: basic jsonl file logger (essentially same as `async` but synchronous)
- `tensorboard`: Uses PyTorch's `torch.utils.tensorboard.SummaryWriter` to write logs in tensorboard format.
- `wandb`: logs to wandb. Currently untested (I don't have an account and need to set that up first)

User can also specify `--run_name` as a string. Instances of `{rank}`, `{local_rank}`, and `{time}` in the string will be replaced with their respective value.

e.g. `{time}_rank{rank}` -> `2025-04-25T17:26:01.477437_rank0`


src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+            "{local_rank}": os.environ.get("LOCAL_RANK", 0),
+        }
+        for placeholder_pat, value in substitutions.items():
+            run_name = run_name.replace(placeholder_pat, str(value))
+
+        self.run_name = run_name
+        self.log_dir = log_dir
+
+    def setup(self):
+        pass
```
RobotSail, line 62: You should raise `NotImplementedError` here so inheriting classes are forced to implement this.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+    return _flattened
+
+
+class BaseLogger:
+    def __init__(self, run_name: str | None, log_dir: str | Path):
+        if run_name is None:
+            # default template
+            run_name = "{time}_rank{rank}"
+
+        substitutions = {
```
RobotSail, line 50: This should probably be moved out of this method and into either a static method or external function. So you would instead have

```py
run_name = BaseLogger.substitute_placeholders(run_name)
```

And you could extend them as needed.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+        pass
+
+    def log_dict(self, d: dict, step: int | None = None):
+        pass
+
+    def teardown(self, exit_code: int = 0):
+        pass
+
+
+class FileLogger(BaseLogger):
```
RobotSail, line 71: Where is the `__init__` method?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+
+        self.file = open(filename, "a")  # pylint: disable=R1732
+
+    def log_dict(self, d: dict, step: int | None = None):
+        if step is not None and not "step" in d:
+            d["step"] = step
+
+        self.file.write(json.dumps(d, indent=None) + "\n")
+
+    def teardown(self, exit_code: int = 0):
```
RobotSail, line 84: Better to indicate these are unused

```suggestion
    def teardown(self, **unused_kwargs):
```

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+
+class WandbLogger(BaseLogger):
+    def __init__(self, run_name: str | None, log_dir: str | Path):
+        super().__init__(run_name, log_dir)
+
+        if wandb is None:
+            msg = (
+                "Could not initialize WandbLogger because package wandb could not be imported. \
+                \nPlease ensure it is installed by running 'pip install wandb'"
+            )
```
RobotSail, line 116: To make this more readable, you can make this multiline instead of escaping with `\` 

```suggestion
            msg = (
                "Could not initialize WandbLogger because package wandb could not be imported.\n"
                "Please ensure it is installed by running 'pip install wandb'"
            )
```

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+    def teardown(self, exit_code: int = 0):
+        self._run.finish(exit_code=exit_code)
+
+
+class TensorBoardLogger(BaseLogger):
+    def __init__(self, run_name: str | None, log_dir: str | Path):
+        super().__init__(run_name, log_dir)
+
+        if SummaryWriter is None:
+            msg = (
```
RobotSail, line 136: same thing here

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+        log_dir = Path(self.log_dir) / self.run_name
+        os.makedirs(log_dir, exist_ok=True)
+
+        self.writer = SummaryWriter(log_dir=str(log_dir))
+
+    def log_dict(self, d: dict, step: int | None = None):
+        flat_dict = flatten_dict(d, sep="/")
+
+        for k, v in flat_dict.items():
+            self.writer.add_scalar(k, v, global_step=step)
```
RobotSail, line 152: One thing I would look out for here is that all values from the existing logs can be serialized into SummaryWriter

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def teardown(self, exit_code: int = 0):
+        self.writer.close()
+
+
+LOGGERS: dict[str, type[BaseLogger]] = {
```
RobotSail, line 158: Do you need `type[BaseLogger]` here? I think just `BaseLogger` might be fine

src/instructlab/training/main_ds.py
```diff
@@ -653,7 +653,8 @@ def main(args):
                 "avg_samples_per_batch": len(dataset) / len(train_loader),
                 "samples_per_gpu": args.samples_per_gpu,
                 "total_samples": len(dataset),  # emit the total number of samples
-            }
+            },
+            step=None,
```
RobotSail, line 657: This log step is a little bit different than the logs in training in that it's mainly just to emit what information training was launched with. Have you investigated what happens to TensorBoard and Wandb when this info gets logged? We want to make sure that it doesn't mess up the format somehow.

RobotSail (changes requested):
Thanks for the PR! A few things need to be fixed, but overall looks good!

src/instructlab/training/logger.py
fynnsu: The linter was complaining about an "unnecessary" init method because it only called `super().__init__()`. I can add it back in and get the linter to ignore it?

RobotSail: No I think it should be fine to leave as-is then.

src/instructlab/training/main_ds.py
fynnsu: Yes, I'll verify that this gets displayed in an okay way. Otherwise could have different logic when `step==None` or perhaps a second method specifically for logging config related stuff.

RobotSail: Yeah I _think_ for TensorBoard it should be fine. I've only worked with Wandb a little bit so that's the one where I would want to try.

src/instructlab/training/logger.py
RobotSail: I would double-check with MyPy, but if that works then let's keep it as-is.

src/instructlab/training/main_ds.py
fynnsu: Yeah, tensorboard didn't display it particularly nicely so I've added a "log_hparams" method, which defaults to just calling `log_dict` but has provides special handling for tensorboard and wandb.

RobotSail: Sweet, were you able to confirm with Wandb also?

src/instructlab/training/main_ds.py
```diff
@@ -930,6 +931,8 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         help="Save full model state using Accelerate after finishing an epoch.",
     )
     parser.add_argument("--log_level", type=str, default="INFO")
+    parser.add_argument("--run_name", type=str, default=None)
+    parser.add_argument("--logger_type", type=str, default="async")
```
booxter, line 935: you can list allowed choices: https://docs.python.org/3/library/argparse.html#choices

booxter (commented):
Main points:

1. Consider using logging module for file management (only use a custom Formatter for JSONL).
2. Define and enforce the form of the input dicts ("a recursive dict of string values")

We will also need some unit tests to validate the new addition. Overall, this looks like a very good start. Kudos.

src/instructlab/training/main_ds.py
fynnsu: Just completed a test on wandb. Loss curve is logged identically to tensorboard, all metrics came through, and the hyper parameters are displayed correctly in the run overview section.

runs/2025-04-28T11:53:34.123994_rank0.jsonl
```diff
@@ -0,0 +1 @@
+{"script_params": {"model_name_or_path": null, "data_path": null, "output_dir": "runs", "num_epochs": 1, "current_epoch": 0, "last_step": 0, "effective_batch_size": 3840, "learning_rate": 0.0001, "lr_scheduler": "cosine", "num_warmup_steps": 1000, "save_samples": null, "save_samples_ds": null, "save_last": false, "checkpoint_at_epoch": false, "accelerate_full_state_at_epoch": false, "log_level": "INFO", "run_name": null, "logger_type": "file", "seed": 42, "mock_data": false, "mock_len": 2600, "distributed_training_framework": "fsdp", "fsdp_sharding_strategy": "HYBRID_SHARD", "use_dolomite": false, "lora_r": 0, "lora_alpha": 32, "lora_dropout": 0.1, "lora_quant_bits": null, "lora_target_modules": null, "max_batch_len": 60000, "cpu_offload_optimizer": false, "cpu_offload_params_fsdp": false, "cpu_offload_optimizer_pin_memory": false, "cpu_offload_optimizer_ratio": 1.0, "NEFTune_alpha": null, "chat_tmpl_path": null, "disable_flash_attn": false, "keep_last_checkpoint_only": false, "use_liger": false}}
```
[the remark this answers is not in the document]
fynnsu: Sorry, I rebased and accidentally committed that. Should be gone now.
Also, I'm not adding it to the `.gitignore` because it's just a random output dir I used during testing, and not specified in the code at all (e.g. as a default value for output_dir).

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+# Standard
+from collections.abc import Mapping
+from datetime import datetime
+from pathlib import Path
+import json
+import os
+
+try:
+    # Third Party
+    import wandb
```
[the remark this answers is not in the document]
fynnsu: I'm not sure what is best here, so will defer to the team. However, my reasons for not including it are:
1. We will potentially have support for 3-4 different logging libraries and would then need an optional dependency for each.
2. Each of these dependency groups would be one package each. It isn't necessarily easier or more logical to do `pip install instructlab-training[wandb]` than it is to do `pip install instructlab-training wandb`
3. It should be relatively clear to the user that they need to install `wandb` to use the WandbLogger and if not the error message should clarify that.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+except ImportError:
+    wandb = None
+
+try:
+    # Third Party
+    from torch.utils.tensorboard import SummaryWriter
```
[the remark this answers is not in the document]
fynnsu: It's the `tensorboard` package.

src/instructlab/training/main_ds.py
RobotSail: Awesome 🔥

booxter:
One other scenario that may be useful is being able to use multiple log destinations at the same time. This would allow to collect same metrics in multiple formats and use preferred tools for different analysis. This is where integration with python logging module could be also helpful since it allows to define multiple destinations for the same messages through [propagate](https://docs.python.org/3/library/logging.html#logging.Logger.propagate).

I'm thinking of enabling all these loggers in CI training runs and collecting all of the outputs as github artifacts.

src/instructlab/training/main_ds.py
RobotSail: In practice it's better to avoid this though, I find choices to be super clunky and not as easy to work with.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+    prefix: str = "",
+    _flattened: dict | None = None,
+) -> dict:
+    """Flatten a nested dictionary into a single-level dictionary.
+
+    Args:
+        d: The dictionary to flatten
+        sep: Separator to use between nested keys
+        prefix: Prefix to add to all keys
+        _flattened: Internal parameter for recursion, should not be set by caller
```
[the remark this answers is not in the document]
fynnsu: Yeah, makes sense I'll update the function to use `|=`.

`prefix` is not currently used but seemed like a useful option to expose because it allows the caller to add a prefix to all elements in the nested dictionary (e.g. `train/` which wandb/tensorboard would use to group items).

src/instructlab/training/logger.py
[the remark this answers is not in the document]
RobotSail: Either way is fine. For PyTorch you still have to install tensorboard separately even though it's technically part of the API. So I don't know if we need a one-off optional requirement just for wandb, but if there are other packages that we need to install to make it work then it could make sense.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+except ImportError:
+    SummaryWriter = None
+
+# First Party
+from instructlab.training import async_logger
+
+
+def flatten_dict(
+    d: Mapping,
+    sep: str = ".",
```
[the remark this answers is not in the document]
fynnsu: Yeah, I was initially using `.` as a separator for some of the loggers, but eventually changed both tensorboard and wandb to `/`. Will update the default.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+        Args:
+            d: Dictionary of values to log
+            step: Optional step number for the log entry
+        """
+        raise NotImplementedError
+
+    def log_hparams(self, hparam_dict: dict, step: int | None = None):
+        """Log hyperparameters.
+
+        Note: calling this method multiple times may overwrite the previous values with matching keys.
```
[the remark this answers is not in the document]
fynnsu: They're not in the base class implementation. However, wandb (and maybe tensorboard) handle them specially and in Wandb's case they are used to set values in a config dictionary, which will be overwritten if the same keys are passed twice.

There reason I added this note here is because our library code will be working with a general "logger" which could end up being a wandb logger or just a basic jsonl file logger. So in the general case we don't want this method to be called multiple times with the same keys.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+        for placeholder_pat, value in substitutions.items():
+            run_name = run_name.replace(placeholder_pat, str(value))
+
+        return run_name
+
+    def setup(self):
+        """Initialize the logger. Must be implemented by subclasses."""
+        raise NotImplementedError
+
+    def log_dict(self, d: dict, step: int | None = None):
```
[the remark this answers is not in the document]
fynnsu: Yes, definitely. I'll add that to the docstrings.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+    # Third Party
+    from torch.utils.tensorboard import SummaryWriter
+except ImportError:
+    SummaryWriter = None
+
+# First Party
+from instructlab.training import async_logger
+
+
+def flatten_dict(
```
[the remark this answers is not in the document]
RobotSail: Good point, we made this mistake early on and our internal utils ended up being imported in `instructlab/instructlab` which caused all kinds of fun for us in e2e tests.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+
+        Raises:
+            RuntimeError: If wandb package is not installed
+        """
+        super().__init__(run_name, log_dir)
+
+        if wandb is None:
+            msg = (
+                "Could not initialize WandbLogger because package wandb could not be imported.\n"
+                "Please ensure it is installed by running 'pip install wandb'"
```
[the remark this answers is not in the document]
RobotSail: I would push back on this. Recommending users to `pip install` something is a very user-friendly pattern that will apply to 95% of users. On the off-chance they are using something aside from pip, they will know which package manager to use. We should keep this message as-is.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+        raise NotImplementedError
+
+
+class FileLogger(BaseLogger):
+    """Logger that writes metrics to a JSONL file."""
+
+    def setup(self):
+        """Setup the file logger."""
+        filename = Path(self.log_dir) / f"{self.run_name}.jsonl"
+        os.makedirs(filename.parent, exist_ok=True)
```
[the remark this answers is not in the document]
fynnsu: Yes, although currently `log_dir` isn't necessarily a `Path` (could be a string). 

Maybe it makes sense to convert it to a path and canonicalize it before passing to `makedirs`?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+
+        self.file = open(filename, "a")  # pylint: disable=R1732
```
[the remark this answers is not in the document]
fynnsu: Yeah I can look into that. This logger was mostly intended as just the bare minimum logger for initial testing purposes. Its output is almost identical to the Async Structured Logger which I added support for later.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+        """Write a dictionary to the JSONL file.
+
+        Args:
+            d: Dictionary to log
+            step: Optional step number to include in the log entry
+        """
+        if step is not None and not "step" in d:
+            d["step"] = step
+
+        self.file.write(json.dumps(d, indent=None) + "\n")
```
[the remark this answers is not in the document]
fynnsu: Will look into using `logging` to automate this.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        flat_dict = flatten_dict(d, sep="/")
+
+        for k, v in flat_dict.items():
+            try:
+                # Check that `v` can be converted to float
+                float(v)
+            except ValueError:
+                # Occurs for strings that can do not represent floats (e.g. "3.2.3") and aren't "inf" or "nan"
+                self.writer.add_text(k, v, global_step=step)
```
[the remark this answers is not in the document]
fynnsu: Anything that isn't a string or `float`-y will raise a `TypeError` instead of a `ValueError`. `str` is special because they support "3.14", "inf", "nan", etc. I think allowing the `TypeError` to be raised uncaught is reasonable behavior (and I will update the docstrings to better document the type expectations).

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,346 @@
...
+) -> BaseLogger:
+    """Create and initialize a logger of the specified type.
+
+    Args:
+        logger_type: Type of logger to create (must be one of ["file", "wandb", "tensorboard", "async"])
+        run_name: Name of the run
+        log_dir: Directory where logs should be stored
+
+    Returns:
+        An initialized logger instance with `setup()` called
```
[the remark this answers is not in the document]
fynnsu: I added this because in theory the loggers themselves will be part of the public api, and if users instantiate a logger directly, they will need to call `setup()`. Therefore, they may expect they need to call `setup()` on the `logger` returned by this helper function. And currently calling `setup` twice is undefined behavior.

Some options I see are:
1. Merge the `setup()` logic into `__init__`/ call `setup()` from `__init__`. I believe there could be use cases where it is better to be able to separate these two steps but currently they're happening consecutively and perhaps it's unnecessary to separate them at this stage.
2. Make `setup` idempotent. The (small) downside is this has to be done per logger/maintained if new loggers are added.
3. Document the behavior (better).

fynnsu:
> One other scenario that may be useful is being able to use multiple log destinations at the same time. This would allow to collect same metrics in multiple formats and use preferred tools for different analysis. This is where integration with python logging module could be also helpful since it allows to define multiple destinations for the same messages through [propagate](https://docs.python.org/3/library/logging.html#logging.Logger.propagate).
> 
> I'm thinking of enabling all these loggers in CI training runs and collecting all of the outputs as github artifacts.

Yeah that's something I've been thinking about. It would also be very easy to just implement a "MultiLogger" class that just loops through its nested loggers. I will look into the logging module more and try to see if it would work well with wandb/tensorboard. I do think it is useful to have the "metric logger" separate from the regular run logging, so I would want to make sure it's possible to do that while using the logging module for regular logs.

RobotSail:
@fynnsu Yes please do that if you can. I would keep the implementation simple (re-use the existing code you've already written ) and just make it so it does the existing AsyncStructuredLogger as a default and then everything else can just be added on after-the-fact. This way we can still retain log data even when using TensorBoard or anything else.

src/instructlab/training/main_ds.py
booxter: Can you please elaborate? As long as choices are calculated (from a enum or dict keys), one doesn't need to touch the argument at all.

src/instructlab/training/logger.py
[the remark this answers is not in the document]
booxter: Yes please document / enforce input types. I'd suggest to have some common enforcement layer for input dicts that would make sure whatever the particular implementation sees is already "normalized" type-wise.

src/instructlab/training/logger.py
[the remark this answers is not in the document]
booxter: 4. Perhaps this is not a very useful interface and callers can just setup() themselves.

src/instructlab/training/logger.py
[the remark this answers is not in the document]
booxter: I'd force a single type for input parameters. Which would be Path. Makes for simpler implementations and give us some freebies from stdlib.

src/instructlab/training/logger.py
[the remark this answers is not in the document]
booxter: Ah I see. Maybe add to the note that you specifically mean that some implementations may behave like that.

booxter (commented):
This is very nice. Let's tighten the flatten_dict type; add unit tests for logger module; and some documentation / examples explaining how this should be used, and I think this is ready.

src/instructlab/training/data_process.py
```diff
@@ -304,7 +310,7 @@ def process_messages_into_input_ids_with_chat_template(args: DataProcessArgs):
             "The provided dataset is empty, please make sure that your dataset contains samples and try again."
         )
 
-    print(f"\033[92mtokenizing the dataset with {args.model_path} tokenizer...\033[0m")
+    logger.info("Tokenizing the dataset with %s tokenizer...", args.model_path)
```
RobotSail, line 313: Can we please keep the colors?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,471 @@
...
+    # First Party
+    import wandb
+except ImportError:
+    wandb = None  # type: ignore
+
+# Third Party
+from rich.logging import RichHandler
+import torch
+
+# Disable package logging by default
```
RobotSail, line 32: But why? Maybe we can make this configurable through an env?

RobotSail (approved):
This is super cool, awesome changes!

src/instructlab/training/utils.py
```diff
@@ -254,16 +255,10 @@ def pad_collate_fn(batch):
                 [(x["labels"] != -100).sum().item() for x in batch]
             )
 
-            print(
```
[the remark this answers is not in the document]
fynnsu: Yeah, the collate fn is run (I believe) by the data loaders on many threads at once. I had initially changed this to a logging statement but then noticed training would occasionally hang waiting for the data loader, which I believe could be some kind of race condition for the logging (even though in theory python stdlib logging is thread-safe).

I removed these statements but added the `total_length` to the dictionary we return and we are logging these values in `main_ds.py:656`. Admittedly, this only logs the `rank==0` values, but it would be easy to log the others as well. 

I can mention in logging documentation that ideally logging should be done from main thread only.

src/instructlab/training/main_ds.py
```diff
@@ -521,6 +532,8 @@ def train(
                 hf_format=True,
                 epoch=epoch,
             )
+        if local_rank == 0:
```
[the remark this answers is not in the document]
fynnsu: Yeah, I added this because I changed the tqdm progress bar to use the `tqdm.rich` version instead and that will actually fail if you open a second progress bar without closing the first.

But I can also change it back to the original `tqdm` bar and then make a second pr improving the bar. I initially changed it because it worked better/was more consistent with the logging, but you're right that it doesn't really fall in scope for this pr.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,471 @@
...
+    A step can be specified by passing `extra={"step": <step>}` to the logging method (e.g. `logger.info(..., extra={"step": 10)}`).
+    To log hyperparameters, pass a (nested) mapping of hyperparameters to the logging method and set `extra={"hparams": True}` (e.g. `logger.info({"lr": 0.001, "batch_size": 128}, extra={"hparams": True})`).
+    """
+
+    def __init__(
+        self,
+        level: int = logging.INFO,
+        run_name: str | None = None,
+        log_dir: str | os.PathLike = "logs",
+        **tboard_init_kwargs: Any,
```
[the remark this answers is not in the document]
fynnsu: This is actually a quirk of python typing. The type for `*args` and `**kwargs` is the type of an individual item, not the resulting variable.

https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

src/instructlab/training/data_process.py
fynnsu: As discussed on slack, we should still have colors when outputting to console using the `RichHandler`.

src/instructlab/training/logger.py
fynnsu: I'm making it configurable. As to why, I believe this is standard practice for libraries so that they don't pollute application code's logs.

tests/unit/test_logger.py
```diff
@@ -0,0 +1,138 @@
...
+    flat = _flatten_dict(nested, sep=".")
+    assert flat == {"a": 1, "b.c": 2, "b.d.e": 3}
+
+    # Test with prefix
+    flat = _flatten_dict(nested, prefix="test_")
+    assert flat == {"test_a": 1, "test_b/c": 2, "test_b/d/e": 3}
+
+
+def test_substitute_placeholders():
+    # Note: time testing could be done better with freezegun library if added to dev dependencies
```
booxter, line 37: remove

tests/unit/test_logger.py
```diff
@@ -0,0 +1,138 @@
...
+        # Test with partial placeholders
+        before_time = datetime.now()
+        name = _substitute_placeholders("run_{time}")
+        after_time = datetime.now()
+
+        time_str = name.split("run_")[1]
+        parsed_time = datetime.fromisoformat(time_str)
+        assert before_time <= parsed_time <= after_time
+
+        name = _substitute_placeholders("rank{rank}")
```
booxter, line 76: these two checks below don't seem to add anything to what is tested in line 55+; remove?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,829 @@
...
+    This handler expects a (nested) dictionary of metrics or text to be logged with string keys.
+    A step can be specified by passing `extra={"step": <step>}` to the logging method.
+    To log hyperparameters, pass a (nested) mapping of hyperparameters to the logging method
+    and set `extra={"hparams": True}`.
+
+    Features:
+        - Automatic flattening of nested metric dictionaries
+        - Support for scalar metrics and text logging
+        - Hyperparameter logging
+        - Automatic directory creation
```
booxter, line 355: "Automatic directory creation" feature? Wow! :)

---

Is it AI slop? I find these docstrings over the top. Please leave just useful stuff, my hand is tired of scrolling. Code is for humans to read; especially docstrings.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,829 @@
+"""Logging utilities for the InstructLab training module.
+
+This module provides a comprehensive logging system for training machine learning models,
+supporting multiple logging backends including TensorBoard, Weights & Biases (wandb),
+and structured JSONL logging. It includes utilities for handling distributed training
+scenarios and formatting metrics for different logging systems.
+
+Key Features:
```
booxter, line 8: are we going to list all features here?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,829 @@
...
+
+if not os.getenv("INSTRUCT_TRAINING_LOGS"):
+    # Disable package logging by default
+    package_logger = logging.getLogger("instructlab.training")
+    package_logger.addHandler(logging.NullHandler())
+    package_logger.propagate = False
+
+### Helper functions
+
+LogDict = Mapping[str, Union[str, int, float, "LogDict"]]
```
booxter, line 95:
```
from __future__ import annotations

type LogDict = dict[str, str | ... | LogDict]
```

to avoid string type form?

booxter (commented):
This is mostly ready. One item that is missing is: test coverage for wandb and tensorboard. I think we should utilize them in CI, somehow, to avoid breaking them at the very least. What's the plan for these? Do you have anything in mind?

src/instructlab/training/logger.py
fynnsu: Yeah, I'll clean up some of the slop.

src/instructlab/training/logger.py
fynnsu: The `type ... = ...` alias syntax was only introduced in py3.12 and we are currently supporting py3.11 too. With the `type` alias syntax it actually even works without `from __future__ import annotations` although it's possible that they turn that on by default in py3.12 anyways.

Unfortunately with py3.11 it doesn't work, and if I remove the leading `type`, I get a `NameError: name 'LogDict' is not defined`.


I believe `from __future__ import annotations` will apply to types in fn definitions, but I think the way `str | ... |` works is actually through overriding of the `|` operator for types, which will still fail cause "LogDict" isn't defined.

booxter:
@fynnsu I agree enabling consumption of these artifacts by the team is a separate work item (please report these and link so that we have a trace). All I'm striving for here is not **consumption** story but basic **regression testing**: just enable them / trigger in CI (unit tests? or perhaps - also? - enable all backend in smoke runs just so that we know this code is being triggered and doesn't crash hard, even if nothing useful for the team comes out of it, yet.) (I hate merging code that is not run anywhere. It's a path to rot.)

tests/unit/test_logger.py
```diff
@@ -64,21 +58,6 @@ def test_substitute_placeholders():
         name = _substitute_placeholders("test_run")
         assert name == "test_run"
 
-        # Test with partial placeholders
```
booxter: I think lines 68-74 were actually reasonable; just the last two checks 76-80 seem redundant.

src/instructlab/training/logger.py
booxter: Thanks for the explanation.

fynnsu:
@booxter I've added tests for the logging handlers that just check if they run without crashing and produce files/folders in the expected location. I also have it running on all "paths" for the handlers (i.e. specifying a `step`, or logging values as `hparams`).

tests/unit/test_logger.py
fynnsu: Well we actually test the datetime works above twice, so I think it's okay to remove here.

src/instructlab/training/multipack_sampler.py
```diff
@@ -388,8 +389,9 @@ def generate_batches(self, set_stats=False):
         # remove indices where the entries are longer than batch max length
         indices = indices[self.lengths[indices] <= self.batch_max_length]
         if len(indices) < len(self.lengths):
-            print(
-                f"\033[33mDropping {len(self.lengths) - len(indices)} samples longer than batch_max_length. Ensure that the right max_batch_length is used during data processing.\033[0m"
+            warnings.warn(
```
JamesKunstle, line 392: You use `warnings.warn` sometimes and `logger.warn` other times. They're obviously for slightly different concerns but what's your thinking on this?

fynnsu: Yeah, I probably wasn't perfectly consistent but tried to follow the guidelines from here: https://docs.python.org/3/howto/logging.html#when-to-use-logging

tldr: use `warnings.warn` if it's library code and a problem the client code can fix, `logger.warn` otherwise

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,801 @@
...
+
+        # Setup logging with a single backend
+        setup_metric_logger(
+            loggers="tensorboard",
+            run_name="my_run",
+            output_dir="logs"
+        )
+        ```
+    """
+    if not loggers:
```
JamesKunstle, line 743: maybe raise ValueError instead

tests/unit/test_logger.py
```diff
@@ -0,0 +1,178 @@
...
+    def get_log_files():
+        return [
+            p
+            for p in tmp_path.iterdir()
+            if p.name.startswith("test_run") or p.name.startswith("wandb")
+        ]
+
+    # Test non-mapping content gets filtered and file/folder isn't created
+    logger.info("test non-mapping content")
+    time.sleep(0.01)  # Wait for Async Handler
```
JamesKunstle, line 163: would there be a way to synchronously check that logging was successful so we don't rely on a sleep? that seems slightly flaky to me.

JamesKunstle: alternatively, could we emulate this in a fixture?

src/instructlab/training/main_ds.py
```diff
@@ -76,13 +82,15 @@
...
-    setup_logger,
 )
 import instructlab.training.data_process as dp
 
+logger = logging.getLogger("instructlab.training")
+
 
 def setup_optimizer(args, model):
     if args.distributed_training_framework == DistributedBackend.FSDP.value:
+        logger.info("Using AdamW optimizer")
```
JamesKunstle, line 93: This applies elsewhere as well, but for debugging we might want to also directly log the parameters we're instantiating the optimizer / other objects with. This would help in tasks when we're making sure that our UI matches with the downstream object behavior (i.e. making sure everything is wired up correctly.

Not necessarily something we need to do in this PR but something that'd be nice to have in general.

src/instructlab/training/main_ds.py
```diff
@@ -326,11 +341,8 @@ def maybe_resume_training(args, model):
 
             last_step = samples_seen // args.effective_batch_size
             args.__dict__["last_step"] = last_step
-        (
-            print(f"\033[93mStarting from: {last_step}\033[0m")
-            if local_rank == 0
-            else None
-        )
+        if local_rank == 0:
```
JamesKunstle, line 344: could we use the Rank0 filter here?

JamesKunstle (changes requested):
I've looked through everything- it seems to meet our needs very well. I especially like that the metric logging has multiple backends and that filtering is programmatic.

Logging isn't something that I'm especially familiar with, and I'd guess that's the same for most people, so it'd be nice to also have a very short "guide" document to how we're using the `logging` library in the `instructlab-training/docs` folder. It shouldn't be implementation-specific for our lib, but more of a tl;dr for anyone looking to quickly learn and edit our logging system.

src/instructlab/training/logger.py
fynnsu: Yeah the reason for this is because we're directly passing the value from the cli into this function and there needs to be a way to select none. But I can move this check outside the function

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,801 @@
...
+        "{rank}": os.environ.get("RANK", 0),
+        "{local_rank}": os.environ.get("LOCAL_RANK", 0),
+    }
+    for placeholder_pat, value in substitutions.items():
+        run_name = run_name.replace(placeholder_pat, str(value))
+
+    return run_name
+
+
+def _flatten_dict(d: LogDict, sep: str = "/", prefix: str = "") -> dict:
```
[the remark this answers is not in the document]
fynnsu: I agree that `d` is a bad name but going to push back on `sep`->`separator`. I think `sep` is clear and used frequently for this purpose (most notably in python's built-in `print` fn).

tests/unit/test_logger.py
fynnsu: I agree and don't love the current approach, however, it isn't obvious to me how we can test this in a better way.

Also, from my understanding, calling `time.sleep` explicitly sleeps **just the current thread**, which means the interpreter looks for something else to run and the logging thread is available for that purpose so will get started. There's still a chance it won't finish in time before the main thread wakes up and interrupts it but it seems unlikely given it doesn't have to write much.

And empirically, I ran this test 10 times in a row with `time.sleep(0.001)` and it passed every time. So with an order of magnitude longer sleep time, I think it's very unlikely to fail. 

And of course, if we find it is flaky in the future I'd fix it but that would probably involve a larger change in how the async logic is implemented, which feels out of scope for this pr.

src/instructlab/training/main_ds.py
fynnsu: Yeah I agree that once we have this logging system merged we should log more stuff throughout the code. However, we may want to do that on top of #498.

src/instructlab/training/main_ds.py
fynnsu: So the `Rank0` filter is added at the `logger` level. Currently, I've set the default root logger config to not use the filter to match existing behavior where some data is printed for all ranks. We could add it to the root logger, but then every logging call would be filtered.

The other reason to still have these checks is to support users who may configure their loggers differently/not use the `Rank0` filter, so that they don't get spammed.

tests/unit/test_logger.py
```diff
@@ -0,0 +1,178 @@
...
+    logger.addHandler(handler)
+
+    return logger
+
+
+@pytest.mark.parametrize(
+    "handler_cls", [TensorBoardHandler, WandbHandler, AsyncStructuredHandler]
+)
+def test_handlers(handler_cls, tmp_path):
+    os.environ["WANDB_MODE"] = "offline"  # Don't publish logs to wandb
```
booxter, line 150: please clean up the variable after the test with

```
with mock.patch.dict(os.environ, clear=True):
```

(Strictly speaking the variable is not needed for other handlers but wandb, so it shouldn't be set for the other variants... but I guess it's easier to write it this way and probably nothing too bad will come out of it...)

tests/unit/test_logger.py
```diff
@@ -0,0 +1,178 @@
...
+    # Test non-dictionary message
+    record = logging.LogRecord("test", logging.INFO, "", 0, "not a dict", None, None)
+    filter_obj.filter(record)
+    assert record.msg == "not a dict"
+
+
+def _setup_logger_and_handler(logger_name, handler_cls, tmp_path):
+    logger = logging.getLogger(logger_name)
+
+    # Remove all existing handlers
```
booxter, line 130: This should be a fixture that would remove them on test exit. Otherwise you leave loggers lingering after the last test case is complete, right?

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,801 @@
...
+    # Third Party
+    import wandb
+except ImportError:
+    wandb = None  # type: ignore
+
+# Third Party
+from rich.logging import RichHandler
+import torch
+
+if not os.getenv("INSTRUCT_TRAINING_LOGS"):
```
booxter, line 80: Is this variable documented? Should it have INSTRUCTLAB_ prefix to match the naming of another variable we introduced recently? (Do we even need it?)

tests/unit/test_logger.py
```diff
@@ -0,0 +1,178 @@
...
+    for handler in list(logger.handlers):
+        logger.removeHandler(handler)
+        handler.close()
+    time.sleep(0.01)  # Wait for Async Handler
```
booxter, line 134: I think this reveals a problem in async handler, namely that it doesn't implement `close()` properly - leaving the original async loop lingering?

tests/unit/test_logger.py
```diff
@@ -0,0 +1,178 @@
...
+    logger = _setup_logger_and_handler(logger_name, handler_cls, tmp_path)
+
+    def get_log_files():
+        return [
+            p
+            for p in tmp_path.iterdir()
+            if p.name.startswith("test_run") or p.name.startswith("wandb")
+        ]
+
+    # Test non-mapping content gets filtered and file/folder isn't created
```
booxter, line 161: nit: prefer "directory" nomenclature since that's what posix and python stdlib use.

tests/unit/test_logger.py
booxter: It would be tricky to prove that something **did not and will not** happen (that the logger was **not** triggered) unless we tap into the module that makes the decision (filters). (Russell's teapot something something.) That would be a burden. I think it's good enough; we may bump to 0.1 just to be extra cautious.

src/instructlab/training/main_ds.py
booxter: (Granted, that we even have argparse as an interface to the library is not optimal and we should probably get rid of it, switching to proper function arguments.)

src/instructlab/training/logger.py
booxter: Disagree on raising since it's a valid input. What it should probably mean - if doesn't already - is that all previously set loggers should be disabled.

booxter (commented):
Overall I think this is good to go. Some extra care with test env cleanup is advised (restore env vars; clean up loggers), and some usage document as requested by James (and me before, though perhaps I wasn't explicit as to what I ask for). Otherwise I'm ready to merge this in.

tests/unit/test_logger.py
fynnsu: Makes sense, I'll update but was there a reason for setting `clear=True`? 

That would clear any existing environment variables which I'd be concerned could have unexpected results if any logging code uses env variables ever.

src/instructlab/training/logger.py
fynnsu: Yes, `INSTRUCT` was a typo. I added this in response to a request from @RobotSail 
https://github.com/instructlab/training/pull/500#discussion_r2072276878

I can also mention it in the logging doc I add.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,801 @@
...
+            _substitute_placeholders(
+                run_name, default_template="training_params_and_metrics_global{rank}"
+            )
+            + ".jsonl"
+        )
+        self.struct_init_kwargs.setdefault("file_name", maybe_file_name)
+        self._struct_logger = None
+
+    def _setup(self):
+        """Initialize the async logger and create the log file."""
```
booxter, line 651: looks like this docstring is wrong and just setting the class up won't create the file. We may as well move the call to setup into __init__ to simplify matters.

src/instructlab/training/logger.py
```diff
@@ -0,0 +1,801 @@
...
+        os.makedirs(Path(self.struct_init_kwargs["file_name"]).parent, exist_ok=True)
```
booxter, line 652: Should this be moved into the Logger class itself - since it implements the lazy semantics already for the file itself - so perhaps we should create the directory lazily too?

tests/unit/test_logger.py
fynnsu: Yes that's correct there is no logic for stopping the async loop right now. The other related issue, is that this can cause the very last logs to not be written to disk if the program stops very abruptly because there is no logic to wait for any remaining logs. I think in reality, neither of this are big problems in the context of the programs the logger is being run in but would still be good to update the `async_logger.py` code at some stage.

For this pr, I just used whatever logic existed in there.

tests/unit/test_logger.py
booxter: Sorry, that was just a copy paste mistake.

src/instructlab/training/logger.py
fynnsu: Yes, I'll update the docstring and (as suggested in your comment below) move the directory setup logic, but otherwise we should keep the `_setup` method as is for two reasons:
1. Consistency with other handlers
2. Lazy creation of the `AsyncStructuredLogger`

Number 2 is important because although `AsyncStructuredLogger` also implements lazy logic for creating a log file, it does immediately launch the thread async thread / use resources. And the way we are currently writing our config, we actually initialize the Handler objects even if they're not attached to a logger. 

We could avoid this by modifying the config to only add handlers if they're going to be used, but the lazy handler approach works well in general.

tests/unit/test_logger.py
fynnsu: #545

tests/unit/test_logger.py
fynnsu: This unfortunately fails if the logging process itself is slow because the task will be removed from the loop when it starts, not when it ends. This can be seen by adding an `await asyncio.sleep(3)` call to `log()` before writing to file, which makes the test sometimes (but not always) fail.

src/instructlab/training/async_logger.py
```diff
@@ -46,8 +45,14 @@ async def log(self, data):
...
     async def _write_logs_to_file(self, data):
         """appends to the log instead of writing the whole log each time"""
-        async with aiofiles.open(self.file_name, "a") as f:
-            await f.write(json.dumps(data, indent=None) + "\n")
+        try:
+            async with aiofiles.open(self.file_name, "a") as f:
+                await f.write(json.dumps(data, indent=None) + "\n")
+        except FileNotFoundError:
+            # Could fail if the parent directory doesn't exist, so create dir and try again
+            await aiofiles.os.makedirs(Path(self.file_name).parent, exist_ok=True)
```
booxter, line 53: you now duplicate the code dumping to file. how about:

```
if self._first:
    await aiofiles.os.makedirs(Path(self.file_name).parent, exist_ok=True)
    self._first = False
async with aiofiles.open(self.file_name, "a") as f:
    ...
```

docs/logging.md
```diff
@@ -0,0 +1,86 @@
...
+
+We also provide the `IsMappingFilter` and `IsRank0Filter` which can be used to filter logs for the above handlers.
+
+Since the above handlers expect dictionary messages, we recommend using the `IsMappingFilter` to ensure that only log messages with dictionaries are processed. In addition, the `IsRank0Filter` can be used to ensure that only the main process logs to the backend systems, which may be necessary for some of the handlers to prevent duplicate logs.
+
+Additionally, we provide the `FormatDictFilter` which is not a true filter, but rather formats a dictionary message into a string for prettier logging. This filter can be used to format dictionary messages before they are passed to stdlib or third party library handlers that expect string messages.
+
+### Setup calling InstructLab as a library
+
+If you are using `instructlab.training` as a library, you may wish to enable logging in your application, and propagate logs from the `instructlab.training` packageto your logging system.
```
booxter, line 72: `packageto` -> `package to`

docs/logging.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Setup for training with InstructLab directly
+
+If you are training with `instructlab.training` directly, you likely don't need to do anything beyond passing in the `run_name`, `logger_type`, and `output_dir` arguments when calling the script. Internally, the training script will call `setup_root_logger()` and `setup_metric_logger()` with a reasonable default configuration.
+
+```bash
+python src/instructlab/training/main_ds.py \
+    ... \
+    --run_name "my_run" \
+    --logger_type "async,tensorboard,wandb" \
```
booxter, line 84: love how easy it is to do multi-backend, or implement another backend. ❤️

fynnsu:
@booxter @JamesKunstle I've added a `docs/logging.md` file that describes both stdlib logging and its integration into `instructlab.training`. Let me know if anything is unclear or needs further explanation.

booxter (approved):
Thank you for pulling this off and addressing all the comments (sometimes misleading!) I like how this functionality integrates with stdlib logging approach. We should strive to be pythonic.

booxter (approved):

PR #508, opened 2025-04-29
fix: set TORCH_NCCL_BLOCKING_WAIT

cdoern:
apparently according to: https://pytorch.org/docs/stable/distributed.html timeout is not honored unless TORCH_NCCL_BLOCKING_WAIT is set to 1

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -3,8 +3,10 @@
 # Standard
 from copy import deepcopy
 from pathlib import Path
+from typing import Union
 import argparse
 import datetime
+import gc
 import math
 import os
 import re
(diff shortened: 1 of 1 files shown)


cdoern:
https://github.com/Mergifyio backport release-v0.10

JamesKunstle:
There were many lines changed in this PR because of indentation so it's kinda hard to know what the _actual_ change is. Could you point me in the right direction?

cdoern:
sure @JamesKunstle !

here is what I added:

1. 
```
    try:
        cached_nccl_blocking_wait: Union[str, None] = os.environ[
            "TORCH_NCCL_BLOCKING_WAIT"
        ]
    except KeyError:
        cached_nccl_blocking_wait = None
    finally:
        os.environ["TORCH_NCCL_BLOCKING_WAIT"] = "1"
   ```
   this sets the NCCL env var to 1 and caches the original value.

I then wrapped the training code in a try: and added 

```
    finally:
        if cached_nccl_blocking_wait is None:
            del os.environ["TORCH_NCCL_BLOCKING_WAIT"]
        else:
            os.environ["TORCH_NCCL_BLOCKING_WAIT"] = cached_nccl_blocking_wait
        torch.cuda.empty_cache()
        gc.collect()
```

which clears the CUDA cache, calls gc.collect, and sets the cached value of the env variable back.

JamesKunstle:
@cdoern The watchdog didn't honor the timeout anyway.

booxter (changes requested):
This should not be merged because:

- timeout patch it tries to fix is going to be reverted.
- disabling watchdog thread has negative implications.
- this is not fixing the CI issue in the first place since the root cause was disc space.

ktdreyer:
@JamesKunstle I think we can close this because we resolved the NCCL timeouts in https://github.com/instructlab/instructlab/issues/3321 . Is that right?

PR #513, opened 2025-04-30
try to do fsdp 2

RobotSail:
Signed-off-by: Oleg Silkin <97077423+RobotSail@users.noreply.github.com>


RobotSail:
Our accelerate version is still old (due to the dolomite dependency) and doesn't support FSDP2.

aldo-at:
Don't use accelerate ;)

RobotSail:
@aldo-at We can perhaps drop accelerate altogether unless there's a reason to keep it that I'm not seeing.

PR #517, opened 2025-05-01
deps: Remove caps on ROCm dependencies

courtneypacheco:
## Overview

We don't need to cap these dependencies at this time. If we do, we should cap them in a `constraints-dev.txt` file like we do in the core repo, but we don't actually need a constraints file if we're not testing upstream.

### DIFF
--- a/requirements-rocm.txt
+++ b/requirements-rocm.txt
@@ -1,3 +1,3 @@
-flash-attn>=2.6.2,<2.7.0
+flash-attn>=2.6.2
 # required for FSDP updates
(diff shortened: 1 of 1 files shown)


booxter (approved):

ktdreyer:
Are we running this through e2e?

courtneypacheco:
@ktdreyer  We don't have any E2E tests that validate training against AMD ROCm accelerators.

booxter:
There are no tests for ROCm upstream. We don't have hardware (atm).

cdoern (approved):

PR #519, opened 2025-05-01
ci: don't trigger pull_request_target job on its own workflow

booxter:
pull_request_target means that the workflow will be pulled from main,
not from the PR under review; so it's a waste to run the workflow in
this case.


ktdreyer (commented):
The reasoning sounds right to me.

danmcp (approved):

ktdreyer (approved):

PR #528, opened 2025-05-05
Enable pylint 'unused-argument' check

### LINKED ISSUE #527: Add unused parameter check to linting
Seems like we're not currently checking for unused parameters in function/method definitions. This is usually a useful check to ensure the apis we're creating make sense / check for accidentally mixed parameters in code.

Pylint has a check for this: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unused-argument.html

fynnsu:
Closes #527

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -250,7 +250,7 @@ def setup_model(
 
         if args.lora_r > 0:
 
-            def make_inputs_require_grad(module, input, output):
+            def make_inputs_require_grad(module, input, output):  # pylint: disable=unused-argument
                 output.requires_grad_(True)
 
             model.get_input_embeddings().register_forward_hook(make_inputs_require_grad)
--- a/src/instructlab/training/utils.py
+++ b/src/instructlab/training/utils.py
@@ -412,7 +412,7 @@ def convert_loss_to_reduce_sum(model, use_dolomite=False):
             output_attentions: Optional[bool] = None,
             output_hidden_states: Optional[bool] = None,
             return_dict: Optional[bool] = None,
-            **deprecated_arguments,
+            **_deprecated_arguments,
         ):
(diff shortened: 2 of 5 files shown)


src/instructlab/training/main_ds.py
```diff
@@ -250,7 +250,7 @@ def setup_model(
 
         if args.lora_r > 0:
 
-            def make_inputs_require_grad(module, input, output):
+            def make_inputs_require_grad(module, input, output):  # pylint: disable=unused-argument
```
booxter, line 253:
```
(_, __, output)
```

will allow you to remove the ignore comment

src/instructlab/training/utils.py
```diff
@@ -636,7 +636,7 @@ def prepare_peft_model(
                 model.enable_input_require_grads()
             else:
 
-                def make_inputs_require_grad(module, input, output):
+                def make_inputs_require_grad(module, input, output):  # pylint: disable=unused-argument
```
booxter, line 639: ditto

booxter (commented):
I assume the reason why some arguments are left untouched is because we expect these functions could be used by library users and so it's unsafe to remove the arguments.

booxter (approved):
On second thought, `(_, __)` is really confusing (though funny).

src/instructlab/training/main_ds.py
fynnsu: `_module, _input` also works. 

The reason I did it like this is because that is the exact hook api pytorch describes, and I was worried there might be a case where the arguments are specified by name, instead of positionally.

Probably unlikely but given I'm not testing this change in an extensive way, I figured this was safest.

booxter: that's a very good reason actually.

JamesKunstle (approved):
Seems reasonable

PR #535, opened 2025-05-06
ci: add 3.12 smoke workflow flavor

booxter:


booxter:
Testing here: https://github.com/instructlab/training/actions/runs/14870170463

.github/workflows/smoke.yaml
```diff
@@ -101,6 +101,9 @@ jobs:
     needs:
       - start-large-ec2-runner
     runs-on: ${{needs.start-large-ec2-runner.outputs.label}}
+    strategy:
```
booxter, line 104: not sure this is correct approach because

- envs will be shared
- variants would run sequentially (?) so tests would take longer

Maybe a separate workflow is needed to achieve this cleanly, though I hate copy pasting it. Unless we write a crusty generator for workflow files to simplify our lives.

booxter: @ktdreyer @courtneypacheco is there any advice CI folks have here? tl;dr I'd like to run same ec2 job for 2+ python flavors, and I'd rather prefer not to copy paste the workflow. Which makes matrix like a good fit, except that these flavors run sequentially due to them sharing same ec2 start/stop jobs.

booxter:
Leaving out EC2 runner reusable action work to a separate issue: https://github.com/instructlab/training/issues/557

booxter:
That the patch is working can be observed here for py311: https://github.com/instructlab/training/actions/runs/15148654764/job/42590466635

For 3.12 flavor, we'll have to merge the patch before we can trigger it in Workflow UI to prove it works. I don't expect issues with it; but if any arise, we'll deal with them.

RobotSail (approved):

booxter:
Not sure why mergify believes there's a ci-failure (as per label). Everything that ran is green (there's not much because pull_request_target workflows can't be validated with regular checks - hence manual execution linked above.

cdoern (approved):

ktdreyer (approved):

ktdreyer:
I'm not sure about the `ci-failure` label, but I think I know why Mergify doesn't merge this. The [Mergify results for this PR](https://github.com/instructlab/training/pull/535/checks?check_run_id=42590034037) shows an unchecked box next to this line: `#review-requested=0`

The problem is that @JamesKunstle is marked as a reviewer and he has not approved this PR.

In the past, developers handle this situation by removing pending reviewers from the PR.

Issue #536, opened 2025-05-06
Speed up flash-attn install in CI by caching the wheel or pulling some binary build, if possible


booxter:
It's taking ~20 minutes to compile just this package. Most other packages are installed nearly instantly. It would be nice to speed smoke tests up if we find a way to avoid compilation on each CI run.

PR #537, opened 2025-05-07
 add and document building flash-attn with ninja

JamesKunstle:
building flash-attn wheel was taking >2hr without ninja and higher parallelism. this change adds ninja as a default requirement for the package (because it needs to be installed _for_ flash-attn and is tiny).

this change also adds a short note about setting build parallelism with the `MAX_JOBS` environment variable to configure the ninja build for flash-attn

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Install the library:
 pip install instructlab-training 
 ```
 
-You can then install the library for development:
+You can also install the library for development:
 
 ```bash
 pip install -e ./training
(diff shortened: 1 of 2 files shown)


booxter (commented):
This is something we **should** tackle in CI and docs consistently. Please refer to what is done in `smoke` workflow now and see if you can reuse it in docs. (Since upstream makes our lives harder in CI, perhaps we should move some of the dirty installation logic from workflows and put it into a script that we would then refer to from both CI instructions and docs?)

Note that in CI, we [preinstall](https://github.com/instructlab/training/blob/e8eb2849078d60dd46dc25ebcdc83a812d0c37ce/.github/workflows/smoke.yaml#L140-L149) `ninja` as well as other build dependencies and it takes around 20-25 minutes to build `flash-attn`. I think it's still an issue that we should solve (by caching wheels or pull from a binary build?), so I [reported](https://github.com/instructlab/training/issues/536) an issue to follow up on this. (I don't know if it's possible due to the fact we don't pin `torch` to the exact version they build their wheels against. But we should explore it.)

---

I have some reservations as to whether this will work without disabling build isolation for `pip`, but I'd need to confirm this indeed doesn't work to be sure. If you could show verbose command output of the installation process following these instructions and that would confirm that `ninja` is indeed available in the isolated build environment for `flash-attn` wheel, it would dissipate my doubts though.

JamesKunstle:
@booxter 
Excellent feedback, thank you for all of your thoughts.
I think you're right- since this is a dependency build problem rather than a runtime dependency it shouldn't go into our direct requirements. Sounds like this is something we should be documenting around- at least, we should be pointing users toward the upstream requirements for building FA for themselves.

booxter:
@JamesKunstle I think we could factor out the code from the workflow that deals with this mess into a script that we could then point users to.

PR #550, opened 2025-05-13
feat: add medium e2e CI job for each PR

cdoern:
This CI job runs the "SDK" of training in conjunction with `ilab data generate`. The goal here is to test convergence on each PR in a timely manner


cdoern:
closing in favor of #551

PR #551, opened 2025-05-13
feat: add medium e2e CI job for each PR

cdoern:
This CI job runs the "SDK" of training in conjunction with `ilab data generate`. The goal here is to test convergence on each PR in a timely manner

### DIFF
--- a/.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
+++ b/.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
@@ -3,15 +3,325 @@
 name: E2E (NVIDIA L40S x4) SDK Test
 
 on:
+  # only run on PRs that touch certain regex paths
+  pull_request:
+    branches:
+      - main
+    paths:
+      # note this should match the merging criteria in 'mergify.yml'
+      - "**.py"
+      - "tox.ini"
+      - "pyproject.toml"
+      - "requirements.txt"
+      - "requirements-dev.txt"
+      - "constraints-dev.txt"
+      - ".github/workflows/e2e-nvidia-l40s-x4-sdk.yaml" # This workflow
   workflow_dispatch:
     inputs:
       pr_or_branch:
         description: 'pull request number or branch name'
         required: true
         default: 'main'
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  TMPDIR: /home/tmp
+
 jobs:
-  noop:
+  start-large-ec2-runner:
     runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.launch-ec2-instance-with-fallback.outputs.label }}
+      ec2-instance-id: ${{ steps.launch-ec2-instance-with-fallback.outputs.ec2-instance-id }}
+      ec2-instance-region: ${{ steps.launch-ec2-instance-with-fallback.outputs.ec2-instance-region }}
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
```diff
@@ -3,15 +3,317 @@
 name: E2E (NVIDIA L40S x4) SDK Test
 
 on:
+  pull_request:
```
JamesKunstle, line 6: needs to be `pull_request_target` to access secrets

.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
```diff
@@ -3,15 +3,317 @@
...
+          ec2_instance_type: g6e.12xlarge
+          aws_resource_tags: >
+            [
+              {"Key": "Name", "Value": "instructlab-ci-github-large-runner"},
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"},
+              {"Key": "GitHubRef", "Value": "${{ github.ref }}"},
+              {"Key": "GitHubPR", "Value": "${{ github.event.number }}"}
+            ]
+
+  e2e-large-test:
```
JamesKunstle, line 87: these are saying large runner, but I think you're starting a medium runner.

JamesKunstle (commented):
I'm not against merging this as a useful test because it's much faster than the large e2e that we use in the upstream `ilab` repo. However, I think it'd be _nicer_ if we called this an `integration-e2e` rather than an `sdk` test. We can isolate the SDK testing slightly differently s.t. we're testing against useful upstream data rather than tightly coupling our testing vs. the SDG output.

.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
JamesKunstle: @cdoern

cdoern: this is our large runner -- l40sx4

.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
```diff
@@ -3,15 +3,317 @@
...
+
+      - name: Fetch and checkout PR
+        if: ${{ github.event_name == 'pull_request_target' }}
+        run: |
+          git fetch origin pull/${{ github.event.number }}/head:pr-${{ github.event.number }}
+          git checkout pr-${{ github.event.number }}
+
+      - name: Update instructlab-training library
+        run: |
+          export CUDA_HOME="/usr/local/cuda"
```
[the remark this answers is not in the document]
cdoern: hmmm, I don't want to install ilab from main though, this is meant to be a step away from having dependencies on ilab main and eventually we don't want any ilab dependency at all. So I'd instead like to maintain our own CUDA installation per repo perhaps? or maybe we should split the action into a `install-cuda` and `install-ilab`?

.github/workflows/e2e-nvidia-l40s-x4-sdk.yml
```diff
@@ -3,15 +3,317 @@
 name: E2E (NVIDIA L40S x4) SDK Test
 
 on:
+  pull_request_target:
+    branches:
+      - "main"
+  schedule:
```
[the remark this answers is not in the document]
cdoern: I can remove schedule

JamesKunstle (approved):

PR #552, opened 2025-05-14
enable fine tuning on HPU

splotnikv:
This PR enables fine tuning on HPU. It supports FSDP, torch.compile and bucketing. HPU is autodetected and the change doesn't affect existing functionality. PR passed basic smoke testing on HPU, more testing is in progress.


src/instructlab/training/multipack_sampler.py
```diff
@@ -211,11 +213,11 @@ def ffd_check_padding(a: np.ndarray, c: int, n: int):
         not_found = True
         for idx in range(n):
             # Calculate the new capacity if size is added to the bin
-            new_capacity = max(bins_max_lengths[idx], size) * (
+            new_capacity = bucket(max(bins_max_lengths[idx], size)) * (
```
RobotSail, line 216: Why are we using `bucket` in multipack? This affects sampling for NVIDIA and AMD systems as well

src/instructlab/training/multipack_sampler.py
```diff
@@ -266,11 +268,11 @@ def ffd_with_result_padding(a: np.ndarray, c: int, start_index: int):
         add_new = True
         for idx in range(len(bins_max_lengths)):
             # Calculate the new capacity if size is added to the bin
-            new_capacity = max(bins_max_lengths[idx], size) * (
+            new_capacity = bucket(max(bins_max_lengths[idx], size)) * (
```
RobotSail, line 271: This shouldn't go in the logic for multipack

RobotSail (changes requested):
Thanks for the PR! The block alignment for Gaudi should happen outside of the Multipack logic. Please update your PR so that samples are correctly aligned for Gaudi systems outside of the sampler logic.

src/instructlab/training/multipack_sampler.py
splotnikv: Multipack decides how many samples and of what size we can fit in any given batch. Bucketing increases maximum sample length of the batch and thus changes batch layout. Moving bucketing out of multipack decreases bucketing efficiency. About impact on other platforms. `bucket()` is defined as "identity" functions on anything except HPU, so should not have any impact.

splotnikv:
> Thanks for the PR! The block alignment for Gaudi should happen outside of the Multipack logic. Please update your PR so that samples are correctly aligned for Gaudi systems outside of the sampler logic.

I've removed bucketing from multi-pack sampler for now, but we have to find a place there to implement it. I've left it in collate function but it is not enough. Please advise how we can do it. E.g. I can implement HPU specific multi-pack sampler.

src/instructlab/training/multipack_sampler.py
splotnikv: Please see my reply above.

src/instructlab/training/multipack_sampler.py
RobotSail: @splotnikv Why do you think that bucketing would be ineffective when moved outside of multipack? From multipack's perspective, it's just looking at lengths. If you move the buckets outside you should still get the same values I believe.

JamesKunstle: Functionally, I'm seeing that `bucket` only returns non-identity if Gaudi is available, so I think @splotnikv is correct that this doesn't impact cuda. 

That being said, re: this implementation, the integration would be preferred as:
```
new_capacity = max(bins_max_lengths[idx], size) ...
if is_torch_hpu_available():
    new_capacity = bucket(new_capacity)
```

rather than obfuscating the cuda path inside of the `bucket` function.

JamesKunstle:
In the next week or so we're going to implement some significantly better testing for this lib that'll give us more confidence in merging this.

My ideal requirements to OK merging this are:
1. test coverage verifying that these modifications don't modify the training behavior on cuda machines.
2. test coverage verifying that supported features work on Gaudi machines (requires that we have a Gaudi runner, a project in-progress).
3. documentation of the bucketing strategy used here and how it interacts with our current multipack impl.

In general, this seems like a pretty minimally invasive integration, so I don't have much hesitation about merging it. I'd just like to make sure we've got good performance / correctness benchmarks ahead of time to hedge against regressions.

src/instructlab/training/multipack_sampler.py
RobotSail: @JamesKunstle Regardless of `bucket`'s behavior on other platforms, Multipack implementation is intended as a generic algorithm which should only operate on what it believes are numbers. 

It would be equivalent and preferred to simply run `bucket` on all of the lengths ahead of time and pass them into Multipack, rather than updating the algorithm to now also consider what platform it's running on.

splotnikv:
@RobotSail, I've refactored bucketing implementation, moved it almost completely out of multipack sampler. One conditional call still remains in `generate_batches()` function, please review.

@JamesKunstle , I've done minimal testing on HPU on our site, everything seems to work as expected, more testing is in progress, I'll put results here when they'll be finished. I am also working on documentation, will add it in next commit.

src/instructlab/training/multipack_sampler.py
```diff
@@ -395,6 +397,11 @@ def generate_batches(self, set_stats=False):
             )
 
         lengths = self.lengths[indices]
+
+        if is_torch_hpu_available():
+            bucket_v = np.vectorize(bucket)
+            lengths = bucket_v(lengths)
```
RobotSail, line 403: @splotnikv This implementation makes more sense, but It seems like we are still mixing responsibilities with what is expected from the Multipack sampler here.

How can we provide the bucketed lengths prior to them even entering the sampler in the first place?

splotnikv: Hi  @RobotSail, unfortunately this is the best that we can do without refactoring a multipack. The problem is in how multipack handles sample lengths. It takes it from two different places (1) from dataset, see `get_effective_samples_per_minibatch()` and (2) from arguments of `MultipackDistributedBatchSampler` ctor. If I remove bucketing from `generate_batches`, I'll have to move it to `get_effective_samples_per_minibatch` and plus add one more call in `setup_dataloader`. I.e. to call bucketing in all places where we create `MultipackDistributedBatchSampler`. Let me know if you prefer this solution. I'll update PR.

splotnikv:
> Hey folks, the refactor is a priority for us right now and we would like to merge this PR in ASAP. So I am going to copy this PR and help you rebase it

@thisisatharva-rh  Hi, I am working on rebase, it does not work yet on Gaudi.

src/instructlab/training/hpu_utils.py
```diff
@@ -0,0 +1,49 @@
...
+def is_torch_hpu_available() -> bool:
+    try:
+        import habana_frameworks.torch.core  # noqa: F401
+    except ImportError:
+        return False
+    return True
+
+
+def simple_bucket(length):
+    """
```
thisisatharva-rh, line 15: Could you explain to me why we need the bucketing algorithm? I see in `main_ds.py` you are setting `lazy_mode=False` which would mean we are using eager compilation, and afaik, eager mode in torch supports dynamically shaped tensors (which I am assuming is the case for habana torch too). I would really appreciate it if you can shed some light on this.

src/instructlab/training/accelerator.py
```diff
@@ -152,14 +161,18 @@ def get_fsdp_config(self):
         prefetch_policy = (
             BackwardPrefetch.BACKWARD_POST if is_lora else BackwardPrefetch.BACKWARD_PRE
         )
-        fsdp_plugin = FullyShardedDataParallelPlugin(
+        fsdp_plugin = (GaudiFullyShardedDataParallelPlugin if is_torch_hpu_available() else FullyShardedDataParallelPlugin)(
```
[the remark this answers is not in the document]
splotnikv: Done in most places. Used existing `device` argument that is supported by cli and config. Please have a look. I'll update what is left in the next commit.

src/instructlab/training/hpu_utils.py
splotnikv: Good question. Even for eager mode we recompile graph if shapes have changed. I'll find out more details and post them here.

braultatgithub:
@RobotSail , what should be done to move this forward?

src/instructlab/training/model.py
```diff
@@ -78,6 +80,15 @@ def __init__(
 
     def _post_model_init(self):
         """Common initialization steps that should happen after model initialization."""
+
+        if self.device == "hpu" and os.getenv("HPU_ENABLE_TORCH_COMPILE", False):
+            cache_size_limit = 10*1000
+            torch._dynamo.config.cache_size_limit = cache_size_limit
+            torch._dynamo.config.accumulated_cache_size_limit = 2*cache_size_limit
+            self.model = torch.compile(self.model, backend="hpu_backend", dynamic=False)
```
bsochack, line 88: Why do you force dynamic=False?

src/instructlab/training/model.py
```diff
@@ -78,6 +80,15 @@ def __init__(
...
+            for layer in self.model.model.layers:
+                layer.compile(backend="hpu_backend", dynamic=False)
```
bsochack, line 90: Why there is per layer compilation?
I think it should be either one or another.
In many cases per layer compilation (regional compilation) is more efficient. It should be selected as an option.

Also same comment, why do you force dynamic=False ?

splotnikv: If we just compile the model, TC skips decoder layers. So, we have to compile them manually. We can do per layer compilation, but in this case `for` loop will look more complicated. We'll have to process `Embedding` layer etc. This is the first part of the model for your reference.
```
GaudiLlamaForCausalLM(
(model): GaudiLlamaModel(
  (embed_tokens): Embedding(32008, 4096)
  (layers): ModuleList(
    (0-31): 32 x GaudiLlamaDecoderLayer(
       (self_attn): GaudiLlamaAttention(
```

If you think that we should add command line option to choose between per model vs per layer compilation, let me know. I'll add. But in both case we will do decoder layer compilation manually.

PR #566, opened 2025-05-22
adds barriers after checkpoint saving

JamesKunstle:
non-rank-0 processes were continuing on and waiting for rank-0 forward, while rank 0 was saving artifacts. This was leading to collective timeouts. In testing, adding these barriers reduced the incidence of timeouts because non-rank-0 processes are waiting at a known wait-point.


RobotSail (approved):
Thank you for this PR, LGTM!

PR #567, opened 2025-05-22
remove old Deepspeed-native code

JamesKunstle:
after moving to Accelerate, we left some old code in-tree in case we had to switch back. We can remove it now that we've settled on Accelerate.


JamesKunstle:
running CI to spot-check that removed code didn't break training. I checked for references to code being used but didn't find any with vscode, but maybe missed something.

RobotSail (commented):
Looks good, if we can get CI passing then I can approve ✅

booxter (approved):

booxter:
@JamesKunstle unless you really need the other two reviews, consider removing them from the requests so that this PR can merge.

JamesKunstle:
Rebased, no other changes

PR #578, opened 2025-05-29
use LigerCrossEntropy loss with reduction=sum

thisisatharva-rh:
Currently, we are using `torch.nn.CrossEntropyLoss(reduction=sum)` with LK to facilitate multipack sampling, which leads to some weird behaviours as reported [here](https://issues.redhat.com/browse/RHELAI-4113). This PR enables us to use `LigerCrossEntropyLoss` with the multi-pack sampler, and solves the incorrect loss issue.

### DIFF
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -55,6 +55,7 @@ from transformers import (
 )
 import torch
 import torch.distributed
+from torch.nn import CrossEntropyLoss
 
 # First Party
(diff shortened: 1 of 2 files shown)


JamesKunstle (commented):
This seems like a reasonable fix.

thisisatharva-rh:
Closing due to some git-tree issues while rebasing, opened [#609](https://github.com/instructlab/training/pull/578) instead

PR #583, opened 2025-05-30
fix: Uncap accelerate

booxter:
Dolomite is removed and we can move forward on this.


RobotSail:
@booxter We cap accelerate because Dolomite + FSDP breaks with newer versions of accelerate. We need to fully deprecate Dolomite before we can raise this dependency.

JamesKunstle:
Dolomite is being deprecated in #589

booxter:
Unclear if introducing such changes at this point in main branch is aligned with org plans for the branch. Feel free to re-open / re-use if needed.

PR #589, opened 2025-06-03
remove GPTDolomite

JamesKunstle:
Dolomite has become a legacy performance and compatibility dependency required before Granite was upstreamed to Huggingface. For >=python3.12, Dolomite has been found to not work correctly. Therefore, we're dropping support.


src/instructlab/training/utils.py
```diff
@@ -706,43 +578,44 @@ def get_module_class_from_name(
...
-            )
-
-        apply_activation_checkpointing(
-            model,
-            checkpoint_wrapper_fn=checkpoint_wrapper_function,
-            check_fn=_whether_to_checkpoint,
-        )
-
-    block_checkpointing(model, **kwargs)
+# # this function is for supporting gradient checkpointing for padding free
```
booxter, line 581: remove?

src/instructlab/training/main_ds.py
```diff
@@ -931,7 +899,11 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         default="HYBRID_SHARD",
         help="Sharding strategy to be used for FSDP distributed training.",
     )
-    parser.add_argument("--use_dolomite", action="store_true")
```
booxter, line 934: Above, I think we can remove 

```
    if train_args.use_dolomite:
        command.append("--use_dolomite")
```

src/instructlab/training/main_ds.py
```diff
@@ -247,28 +235,11 @@ def setup_model(
             model,
             lora_config,
             args.distributed_training_framework,
-            gradient_checkpointing=not args.use_dolomite,
```
booxter, line 250: now that it's always True, I think you could do some code simplification in `prepare_peft_model` by removing the argument and hardcoding the True value inside the function.

src/instructlab/training/main_ds.py
```diff
@@ -247,28 +235,11 @@ def setup_model(
...
+            gradient_checkpointing=True,
         )
         args.lora_config = lora_config
-    elif not args.use_dolomite:
-        model.gradient_checkpointing_enable()
```
booxter, line 254: before the change, in case of `lora_r > 0`, we were not calling `gradient_checkpointing_enable`. Now we do. Is it an intentional change?

src/instructlab/training/utils.py
```diff
@@ -71,14 +57,15 @@ def check_valid_train_args(train_args: TrainingArgs):
             f"Provided path does not exist locally and is not an HF format name. Please make sure that you've passed a valid model path and that it has appropriate permissions, or a Huggingface model name (org/repo): {train_args.model_path}"
         )
 
-    if train_args.use_dolomite and train_args.disable_flash_attn:
-        raise RuntimeError(
-            "ERROR: Trying to use dolomite padding-free transformer without flash attention is not supported"
+    if train_args.use_dolomite:
```
booxter, line 60: consider using https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields instead

src/instructlab/training/utils.py
```diff
@@ -219,224 +204,159 @@ def supports_flash_attention(device_id=0):
     return is_sm8x or is_sm90 or is_compat_amd
 
 
-def check_flash_attn_enabled(disable_flash_attn: bool, use_dolomite: bool) -> bool:
+def check_flash_attn_enabled(disable_flash_attn: bool) -> bool:
     if not disable_flash_attn:
```
booxter, line 208: this code is now quite silly. :) how about

```
if disable_flash_attn:
    return False
if supports...():
    return True
raise ...
```

JamesKunstle:
Will re-open this PR based off of changes from #572

PR #594, opened 2025-06-04
feat: refactor main_ds.py (2/n) Accelerator class

cdoern:
** PLEASE NOTE, THIS PR INCLUDES THE CHANGES IN #572, AND WILL BE REDUCED IN SIZE ONCE THAT MERGES **

Introduce a new design for key components of main_ds.py. Namely splitting Model initialization, Accelerator initialization, Optimizer initialization, and Checkpoint saving initialization into classes. This commit introduces the Accelerator class

The Accelerator class aims to both store commonly accessed variables associated with the accelerated model and abstract model/optimizer mutation away from the user who should only access our Model and Optimizer classes.

Being able to group functionality under the Model, Accelerator, and Checkpointer classes inherently reduces code complexity and duplication. Being able to store things like , self.distributed_framework,self.lora_config, etc in a way such that within the class they are accessible within different methods allows the arguments per method to go down drastically, as well as complex return values. Simpler methods and argument/return values allows for simpler testing of code.


src/instructlab/training/accelerator.py
```diff
@@ -0,0 +1,250 @@
...
+)
+
+# Local
+from .model import Model
+
+
+class Accelerator:
+    def __init__(
+        self,
+        model: Model,
```
thisisatharva-rh, line 24: `Model` acts as a "factory" class that creates `nn.module` once the `from_pretrained` method is called in the case of each; Liger, Dolomite and the normal transformer model. That is what we should pass into the Accelerator so that we can avoid the weird `model.model` references.

thisisatharva-rh: yeah, that's correct. I guess we need to refine the model class more for that to happen; approving rn in the spirit of getting the refactor in quickly.

thisisatharva-rh (approved):
we need to have a larger conversation about what the `Model` class should look like, but for now, this looks good.

booxter (approved):

Issue #592, opened 2025-06-04
`tox -e py3-smoke` is broken because of flash-attn


JamesKunstle:
When running `tox -e py3-smoke`, flash-attn can't be built and test fails to run. `tox_current_env` has to be used. This is bad user experience; a user should be able to expect that running a tox environment target should only fail if their code fails.

Issue #548, opened 2025-05-09
Move code preparing venv with flash-attn from github workflow to a script


booxter:
This code is useful for anyone trying to set up the environment (either another workflow, or anyone working with the library). So we should move it from yaml into a shell script, then refer to the script from the workflow.

https://github.com/instructlab/training/blob/e8eb2849078d60dd46dc25ebcdc83a812d0c37ce/.github/workflows/smoke.yaml#L140-L161

While at it, we should check if we shouldn't reuse the same code in any other (e2e?) jobs.

booxter:
It may be wise to sync with ilab repo since they are also going to do it there: https://github.com/instructlab/instructlab/pull/3399

PR #597, opened 2025-06-04
adds build step for flash-attn to tox py3-smoke env

JamesKunstle:
flash-attn can't be built inline with other dependencies because it requires torch and ninja. we're currently getting around this with a `tox_current_env` setup but that's tricky from a tox perspective. this change makes it so that we can use tox to build the environment directly, and that the environment won't be rebuilt in between command runs.

Closes: #548
Closes: #592

### DIFF
--- a/scripts/install_flash_attn_if_needed.py
+++ b/scripts/install_flash_attn_if_needed.py
@@ -1,0 +1,27 @@
+# Standard
+import importlib.util
+import subprocess
+import sys
+
+
+def flash_attn_is_installed() -> bool:
+    return importlib.util.find_spec("flash_attn") is not None
+
+
+if flash_attn_is_installed():
+    print("flash-attn already installed, skipping.")
+else:
+    print("flash-attn not found in environment, installing...")
+    subprocess.check_call(
+        [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "-r",
+            "requirements-cuda.txt",
+            "-c",
+            "constraints-dev.txt",
+            "--no-build-isolation",
+        ]
+    )
--- a/tox.ini
+++ b/tox.ini
@@ -11,8 +11,8 @@ wheel_build_env = pkg
 deps =
     pytest
 install_command = pip install \
-                  -c constraints-dev.txt \
-                  {opts} {packages}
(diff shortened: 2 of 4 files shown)


JamesKunstle:
Fixes #592

tox.ini
```diff
@@ -11,8 +11,8 @@ wheel_build_env = pkg
 deps =
     pytest
 install_command = pip install \
-                  -c constraints-dev.txt \
```
booxter, line 14: could avoid touching it

booxter (approved):
Thank you! This is a lot cleaner!

booxter:
Huh... https://github.com/instructlab/training/actions/runs/15451490447/job/43494664680?pr=597

scripts/install_flash_attn_if_needed.py
```diff
@@ -0,0 +1,26 @@
...
+    subprocess.check_call(
+        [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "-r",
+            "requirements-cuda.txt",
+            "-c",
+            "constraints-dev.txt",
```
booxter, line 24: based on the failure in smoke run, I assume you still want to pre-install torch and other deps before plunging into flash-attn installation here.

JamesKunstle: Nope still broken... this is the same problem I was seeing before with `torch` not being able to find its cuda dependencies even though it's installed with them

booxter:
@mergifyio rebase

scripts/install_flash_attn_if_needed.py
booxter: Is it because these variables are not allowed by `tox` via `passenv`?

```
        export CUDA_HOME="/usr/local/cuda"
        export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${CUDA_HOME}/lib64:${CUDA_HOME}/extras/CUPTI/lib64"
        export PATH="${PATH}:${CUDA_HOME}/bin"
```

booxter:
@JamesKunstle Do you plan to complete this PR? If not, then maybe close it.

PR #600, opened 2025-06-04
ci: Validate reusable actions with actionlint

booxter:


booxter:
Yeah, actionlint is not doing what I thought it would do. Not sure how to validate reusable actions (not workflows).

PR #604, opened 2025-06-09
chore: Remove hf_format= argument for save_checkpoint

booxter:
It's always True.


booxter:
The change is included in Charlie's refactoring patch, so no need for a separate PR to handle the same.

PR #605, opened 2025-06-10
feat: refactor main_ds.py (3/n) Checkpointer Class

cdoern:
into classes:

1. Model
2. Accelerator
3. Checkpointer

The Checkpointer class introduces a unified approach to our various checkpointing techniques. A user can pass in their checkpointing style (full_state or hf_format), and the checkpointer, via checkpointer.checkpoint, will save the model using the selected method and other techniques (LoRA).

This PR adds the new class and unit tests for the class

see previous PRs #572  and #594 

note: this is probably the last of these large refactor for now with subsequent smaller followup PRs for cleanup.

### DIFF
--- a/src/instructlab/training/checkpointer.py
+++ b/src/instructlab/training/checkpointer.py
@@ -191,7 +191,7 @@ class Checkpointer:
         )
         start = time.time()
 
-        if self.model.model_type in ("gpt_megatron", "gpt_dolomite"):
+        if self.model.model_conf.model_type in ("gpt_megatron", "gpt_dolomite"):
             convert_dolomite = False
         else:
             convert_dolomite = True
@@ -199,7 +199,7 @@ class Checkpointer:
         # Build the final output directory path
         final_output_dir = Path(output_dir) / "hf_format" / subdir
 
-        if self.model.model_type == "dolomite" and convert_dolomite:
+        if self.model.model_conf.model_type == "dolomite" and convert_dolomite:
             tmpdir = TemporaryDirectory("w")  # pylint: disable=consider-using-with
             output_dir = Path(tmpdir.name)
         else:
@@ -236,10 +236,10 @@ class Checkpointer:
             output_dir.mkdir(parents=True, exist_ok=True)
             if not self.model.module.config.architectures and convert_dolomite:
                 arch_added = False
-                if self.model.model_type == "llama":
+                if self.model.model_conf.model_type == "llama":
                     self.model.module.config.architectures = ["LlamaForCausalLM"]
                     arch_added = True
-                elif self.model.model_type == "granite":
+                elif self.model.model_conf.model_type == "granite":
                     self.model.module.config.architectures = ["GraniteForCausalLM"]
                     arch_added = True
                 if arch_added:
@@ -272,7 +272,7 @@ class Checkpointer:
             )
 
         if (
-            self.model.model_type == "dolomite"
+            self.model.model_conf.model_type == "dolomite"
             and convert_dolomite
             and self.accelerator.is_main_process
         ):
@@ -282,7 +282,7 @@ class Checkpointer:
             export_to_huggingface(
                 pretrained_model_name_or_path=tmpdir.name,
                 save_path=final_output_dir,
-                model_type=self.model.model_type,
+                model_type=self.model.model_conf.model_type,
             )
             tmpdir.cleanup()
 
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -362,6 +362,7 @@ def main(args):
         flash_enabled=flash_enabled,
         noise_alpha=args.NEFTune_alpha,
         lora_quant_bits=args.lora_quant_bits,
+        model_conf=model_conf,
     )
 
     args.base_model_args = m.base_model_args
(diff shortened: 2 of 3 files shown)


src/instructlab/training/checkpointer.py
```diff
@@ -0,0 +1,385 @@
...
+    def checkpoint(self, *args, **kwargs):
+        # Calls the method chosen at init
+        return self._checkpoint_fn(*args, **kwargs)
+
+    # pylint: disable=unused-argument
+    def _no_checkpoint(self, *args, **kwargs):
+        print("[None] Skipping checkpointing.")
+
+    # pylint: disable=unused-argument
+    def save_fsdp_lora_model(
```
fynnsu, line 54: Potentially for a future PR, but I think it would be cleaner to have a base `Checkpointer` abstract class and then have `FSDPLoRACheckpointer`, `HFFormatAccelerateCheckpointer`, etc. subclasses which each implement their own `checkpoint` method. Instead of doing our own custom routing with `self._checkpoint_fn`

src/instructlab/training/main_ds.py
```diff
@@ -239,28 +235,20 @@ def train(
...
 
     if args.save_last:
-        save_hf_format_accelerate(
-            args,
-            model,
-            model.tokenizer,
-            accelerator,
-            samples_seen,
-            is_lora=bool(args.lora_r),
+        checkpointer.save_hf_format_accelerate(
```
fynnsu, line 247: Shouldn't this be `checkpointer.checkpoint()`

src/instructlab/training/model.py
```diff
@@ -50,11 +50,13 @@ def __init__(
         flash_enabled: bool = False,
         lora_config: Optional[LoraConfig] = None,
         lora_quant_bits: int = 0,
+        model_conf=None,
```
fynnsu, line 53: What is `model_conf` for? Currently we only seem to be using it to access `model_conf.model_type`. Could we just store `model_type` instead?

Also currently none of those accesses of `model_conf.model_type` check that `model_conf is not None` before trying to access the attribute, so this will raise an error if `model_conf` is ever actually `None` (it's current default value).

src/instructlab/training/main_ds.py
```diff
@@ -483,13 +472,19 @@ def main(args):
...
 
-    load_latest_full_state(args=args, accelerator=accelerator)
-
+    strategy = "all"
+    if not args.accelerate_full_state_at_epoch:
+        strategy = "hf_format"
+    checkpointer = Checkpointer(
+        strategy=strategy, model=m, optimizer=optimizer, accelerator=accelerator
+    )
+    checkpointer.load_latest_full_state(Path(args.output_dir))
```
fynnsu, line 481: Can `args.output_dir` be set in the `Checkpointer.__init__`? It seems like we're currently passing it into every load/checkpoint function but it doesn't seem like it's changing values (or should change values) currently.

fynnsu (commented):
Left some comments below.

I'm also assuming that the contents of the specific methods (like `save_fsdp_lora_model`) is largely unchanged. Is that correct?

src/instructlab/training/checkpointer.py
```diff
@@ -0,0 +1,385 @@
...
+        self,
+        output_dir: Path,
+        **kwargs,
```
booxter, line 57: please remove kwargs that are not used; transform those used into specific arguments for the information that needs to be passed (with proper names, types etc.)

src/instructlab/training/checkpointer.py
```diff
@@ -0,0 +1,385 @@
...
+from .utils import log_rank_0, wraps
+
+
+class Checkpointer:
+    def __init__(
+        self,
+        model: Model,
+        optimizer: torch.optim.Optimizer,
+        accelerator: Accelerator,
+        strategy="all",
```
booxter, line 31: make it enum

src/instructlab/training/checkpointer.py
```diff
@@ -0,0 +1,385 @@
...
+            if old_device_map:
+                # return the previous device_map so it can be used later on if needed
+                self.model.base_model_args["device_map"] = old_device_map
+
+        dist.barrier()
+
+    # pylint: disable=unused-argument
+    def save_full_state(
+        self,
+        output_dir,
```
booxter, line 120: define all args' types.

src/instructlab/training/checkpointer.py
booxter: I made a similar argument for Model class before... Class hierarchies are exactly meant for such scenarios.

tests/unit/test_checkpointer.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@pytest.fixture
+def mock_optimizer():
+    return MagicMock()
+
+
+@pytest.fixture
+def mock_accelerator():
+    accelerator = MagicMock(spec=Accelerator)
```
booxter, line 45: instead of mocks, you could introduce a new subclass for TestAccelerator that would "do nothing" / "do bare minimum" for test purposes. Same for the rest. Why do we have to have mocks just to create an object? Are __init__ methods destructive / invasive? (Maybe it should be fixed then - it should be generally safe / cheap to create objects.)

PR #615, opened 2025-06-18
ci: Use torch CUDA 12.8 index when calculating constraints

booxter:
This should allows to bump to 12.8 (after torch is bumped past 2.6.0).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,9 @@ package = wheel
 wheel_build_env = pkg
 deps =
     pytest
+setenv =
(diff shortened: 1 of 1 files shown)


booxter:
torch bump past 2.6 is here: https://github.com/instructlab/training/pull/614

booxter:
will need https://github.com/instructlab/ci-actions/pull/35 to avoid downgrading of versions in constraints.

booxter:
It's not clear at this point if we want to proceed with newer CUDA for the main branch. I will close and folks can re-open / copy if needed later.

PR #625, opened 2025-06-23
ci: Switch to python 3.12 as default target

booxter:


booxter:
Needs some agreement on version policy: https://github.com/instructlab/dev-docs/pull/200 and then an approval of an update to the doc suggesting 3.12 is the default version...

PR #626, opened 2025-06-23
fix edge case for qwen3 data processing

RobotSail:
With Qwen3, there's an edge case which can result in the unmask/mask logic breaking during data processing.

Root Cause: The error occurs specifically when using the Qwen/Qwen3-32B tokenizer, not with Qwen/Qwen2.5-32B-Instruct. The problematic sample contains multiple <think> tags in the assistant's
 response.

Issue Location: The error occurs in data_process.py:555 in the unmask_messages function, where it encounters an <|UNMASK_END|> token while not in an unmasking state.

Key Findings:
1. Model-specific issue: The sample processes fine with Qwen/Qwen2.5-32B-Instruct but fails with Qwen/Qwen3-32B
2. Chat template differences: Different models have different chat templates that may tokenize the unmask tokens differently
3. Token ordering: The issue suggests that the unmask tokens are getting reordered or processed incorrectly by the Qwen3 chat template

The Problem:
The Qwen/Qwen3-32B model's chat template is processing the <|UNMASK_BEGIN|> and <|UNMASK_END|> tokens in a way that causes them to appear out of order or in an unexpected state, leading to
the algorithm encountering an <|UNMASK_END|> token when it's not actively unmasking.

This is likely due to differences in how the chat templates of Qwen2.5 vs Qwen3 handle special tokens, particularly when there are multiple special tokens or complex content like the <think>
tags present in the assistant's response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ The library now supports an optional `reasoning_content` field in addition to th
 - Supporting models that need to generate internal reasoning traces
 - Enabling step-by-step reasoning in model responses
 
+> **Note**: this is only supported for models with chat templates that use the DeepSeek R1-style parser. Models without a custom thought processor such as Phi-4 must still provide their reasoning traces in the `content` field.
+
 **Example message structure with reasoning content:**
 
 ```json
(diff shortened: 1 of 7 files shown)


Maxusmusti (approved):
LGTM

cdoern:
@mergifyio rebase

PR #637, opened 2025-07-03
update .github from main (commit beae7f14)

ktdreyer:
Our main branch has many CI fixes and improvements that we want to incorporate into our stable release branch. This will increase our confidence in CI results for backport PRs.

Sync our `.github` directory on `release-v0.10` with what is in the `main` branch (commit beae7f14a1ed00a227a88bf961a862e01987528c).

These new tests also require the `constraints-dev.txt` file system. I've copied those changes from `main` as well, and regenerated `constraints-dev.txt`.


ktdreyer:
This uncovered two issues:
1) In `main`, we still do not use [`launch-ec2-runner-with-fallback`](https://github.com/instructlab/ci-actions/blob/main/actions/launch-ec2-runner-with-fallback/launch-ec2-runner-with-fallback.md) on all our tests that use AWS.
2) This change pulls in the latest version of `ruff`, so that test fails now on new issues in the codebase. In the interested of keeping this branch stable and minimizing code changes, we could pin to the older version of ruff in this PR so that ruff will pass with existing code, and then we could cherry-pick the new ruff fixes separately.

PR #654, opened 2025-09-12
Refactors train loop, adds padded batch packer, other fixes

RobotSail:
This PR adds a number of improvements:

1. padded batch packer
2. refactor code so training loop is simpler
3. simplifies batch management and aggregation
4. simplifies gpt-oss saving
5. adds tests for sampler and batch packing algorithms
6. fixes gpt-oss detection during data processing

There are still a few other things to fix but this resolves the bulk of the outstanding items


Maxusmusti (approved):
LGTM

PR #659, opened 2025-09-22
Update requirements-cuda.txt to increase liger-kernel minimum

Maxusmusti:
Found that liger-kernel<0.5.10 was causing some incompatibility issues with more recent models


osilkin98 (approved):
LGTM

PR #660, opened 2025-09-23
Enable fine tuning on HPU

splotnikv:


splotnikv:
I've fixed all "TODOs". PR is ready for review now.

src/instructlab/training/main_ds.py
```diff
@@ -789,8 +810,33 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
         action="store_true",
         help="Use Liger kernels for training.",
     )
+    parser.add_argument(
+        "--device",
```
RobotSail: In the context of PyTorch, we typically assume `device` refers to a singular device on a node containing several accelerators, e.g.: `cuda:0, cuda:1, etc.` 

Since you're looking to allow for switching between HPU and CUDA style training, maybe we could change this to be `--device_type`? This way, we can also constrain the options to be either cuda (by default) or hpu, and error out when given an option which doesn't fall in the set of allowed devices.

src/instructlab/training/sampler.py
```diff
@@ -360,12 +395,14 @@ def get_data_loader(
         num_workers: Number of data loading workers
         flash_enabled: Whether flash attention is enabled (affects collation strategy)
         pad_token_id: Token ID to use for padding (only used when flash_enabled=False)
+        num_chunks: Number of chunks to split dataset into for sequential training
```
RobotSail: Nit: "sequential training" here can be interpreted as training on multiple datasets, but it seems like it's being used to optimize padded batching. Should we clarify that?

src/instructlab/training/model.py
```diff
@@ -57,17 +56,22 @@ def __init__(
...
         self.distributed_framework = distributed_framework
+        self.device = device
+        self.torch_compile = torch_compile
         quant_config = None
 
         # check model type & set on the mclasss
         self.is_gpt_oss = is_gpt_oss(model_path)
         if self.is_gpt_oss:
             # Third Party
+            from transformers import Mxfp4Config  # pylint: disable=no-name-in-module
```
RobotSail: @splotnikv How come this import is being moved from module-level to method-level?

src/instructlab/training/config.py
```diff
@@ -254,3 +254,8 @@ class TrainingArgs(BaseModel):
     log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = Field(
         default="INFO"
     )
+
+    device: Optional[str] = None
+    torch_compile: bool = False
+    num_chunks: int = 1
+
```
RobotSail: I think these are good to expose, but what are your thoughts if we adjusted the field names like this? 

- `device` --> `device_type` (a constrained string of `cuda | hpu`, defaults to `cuda`), to be clearer on what the underlying hardware looks like, rather than a specific device we train on
- `num_chunks` --> `padded_packing_num_sorting_bins` this field represents a very specific implementation detail, so a longer name is fitting. A description should be added to explain that this helps training efficiency when packing has to be used, at the cost of randomness/increased bias. 

Would it also make sense to have it be `enable_torch_compile` or `use_torch_compile`?

RobotSail (commented):
Thank you for the PR @splotnikv , overall this PR seems good and the changes all seem reasonable to enable HPU. I left a few comments inline and just had a few questions:

1. I see that the updated EpochSampler now has the ability to shuffle samples in a way that's biased to be similar in length. Have you compared this approach to having the same algorithm live in the Collator and operate at a batch-level instead? Apologies in advance if this is already documented somewhere in the PR and I missed it. 

2. Would you be ammeanable to using `device_type` instead of `device`? We use both throughout the codebase, so it's important to prevent confusion wherever possible. 

Otherwise, this PR makes sense and would be a great addition. Thank you for your hard work on this!

RobotSail (commented):
Actually, I think it's fine to have the EpochSampler do length-aware sampling. Could you just please update the parameters & docstrings to be clear that the feature is length-biased shuffling, and is a variance-bias tradeoff? Thank you. 

Also: Would you mind updating the docs to showcase your new additions? That way people are aware of the awesome changes you're adding 😃

PR #661, opened 2025-10-03
fix(torchrun): Omit empty arguments and correct nproc_per_node type

szaher:
The command generation logic is updated to dynamically build the torchrun command, excluding arguments that are empty or None. This prevents them from overriding environment variables, ensuring that torchrun can
correctly inherit its configuration. An exception is made for integer arguments where 0 is a valid value.

Additionally, the nproc_per_node argument type has been changed from int to str to support special values
accepted by PyTorch, such as 'auto', 'gpu', and 'cpu'.

Reference: https://github.com/pytorch/pytorch/blob/main/torch/distributed/run.py#L77-L88

### DIFF
--- a/src/instructlab/training/config.py
+++ b/src/instructlab/training/config.py
@@ -75,7 +75,7 @@ class TorchrunArgs(BaseModel):
     https://pytorch.org/docs/stable/elastic/run.html#definitions
     """
 
-    nproc_per_node: int
+    nproc_per_node: str
     nnodes: int
     node_rank: int
     rdzv_id: int
--- a/src/instructlab/training/main_ds.py
+++ b/src/instructlab/training/main_ds.py
@@ -480,26 +480,24 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
     if not os.path.exists(train_args.ckpt_output_dir):
         os.makedirs(train_args.ckpt_output_dir, exist_ok=True)
 
-    command = [
-        "torchrun",
-        f"--nnodes={torch_args.nnodes}",
-        f"--node_rank={torch_args.node_rank}",
-        f"--nproc_per_node={torch_args.nproc_per_node}",
-        f"--rdzv_id={torch_args.rdzv_id}",
-        f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
-        __file__,
-        f"--model_name_or_path={train_args.model_path}",
-        f"--data_path={train_args.data_output_dir}/data.jsonl",
-        f"--output_dir={train_args.ckpt_output_dir}",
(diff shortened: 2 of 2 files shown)


src/instructlab/training/config.py
```diff
@@ -75,7 +75,7 @@ class TorchrunArgs(BaseModel):
     https://pytorch.org/docs/stable/elastic/run.html#definitions
     """
 
-    nproc_per_node: int
+    nproc_per_node: str
```
RobotSail, line 78: Did you mean to make this change?

src/instructlab/training/main_ds.py
```diff
@@ -480,26 +480,24 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
...
+            continue
+        command.append(f"--{key}={value}")
+
+    # append this file as main script to run by torchrun
+    command.append(__file__)
+
+    # build args for this file. Ignore empty or unset values except int values
+    for key, value in train_args.model_dump(exclude_none=True).items():
+        # avoid ignoring int attrs with value = 0
+        if not isinstance(value, int) and (not value or value == ""):
```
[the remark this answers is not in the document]
szaher: I have updated this one to only check for string types.

src/instructlab/training/main_ds.py
```diff
@@ -480,26 +480,24 @@ def run_training(torch_args: TorchrunArgs, train_args: TrainingArgs) -> None:
...
+            continue
+        command.append(f"--{key}={value}")
```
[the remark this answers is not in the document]
szaher: I have updated this one to only process `torchrun` args and leave the scripts args as they're not perfectly 1:1 mapped.

src/instructlab/training/config.py
szaher: yes

src/instructlab/training/config.py
```diff
@@ -75,12 +75,17 @@ class TorchrunArgs(BaseModel):
...
     nnodes: int
     node_rank: int
-    rdzv_id: int
+    rdzv_id: Union[str, int]
     rdzv_endpoint: str
 
+    # this will tell the model construct to ignore
+    # extra arguments that aren't part of this model
+    class Config:
+        extra = "ignore"
```
RobotSail, line 87: @szaher Do you know when this would be the case? If our goal here is to dynamically build the `torchrun` command using the defined interface, this seems like it now opens the floor up for users to pass invalid arguments through torchrun. This means that any incorrect interface usage wouldn't be detected until runtime.

RobotSail: I see, that's fine then.

PR #686, opened 2026-02-26
Add on-demand full-state checkpointing for OpenShift AI / KubeFlow preemption

RobotSail:
Implements signal-driven checkpoint-and-exit for distributed training jobs
running in OpenShift AI as KubeFlow training jobs or multi-node bare metal.

When `on_demand_checkpointing=True` is set in TrainingArgs:

- Parent process (run_training) installs handlers for SIGTERM, SIGINT,
  SIGUSR1, SIGUSR2, SIGXCPU, and SIGHUP — covering all signals
  Kubernetes/OpenShift sends before the hard SIGKILL.
- On signal receipt, a trigger file is atomically written to /dev/shm
  (tmpfs, shared within the pod, zero disk I/O).
- Worker processes check for the trigger file after each optimizer step
  via an all_reduce(MAX) collective, ensuring global consensus across
  all ranks on all nodes.
- When any rank detects the trigger, all ranks collectively save a
  full-state distributed checkpoint (model + optimizer + LR scheduler)
  then exit gracefully.
- Parent waits up to 300s for workers to complete the checkpoint before
  proceeding with normal shutdown.

https://claude.ai/code/session_01HSxsk7SnMULJxy7uafe7t3

## Summary by CodeRabbit

* **New Features**
  * New training option and CLI flag to enable on-demand, signal-triggered full-state checkpointing for preemption.
  * Coordinated parent/worker checkpoint orchestration with signal handling, trigger-file coordination, graceful worker wait, and improved shutdown messaging.
  * Worker-side utilities to detect, save, and clear on-demand checkpoint requests.
  * Batch processing can be interrupted mid-minibatch and now reports an interrupted flag in batch metrics.
  * Mid-epoch checkpoints now record/resume using an optional global-step metadata field.

### DIFF
--- a/docs/on_demand_checkpointing.md
+++ b/docs/on_demand_checkpointing.md
@@ -125,6 +125,52 @@ trigger file, the new run's `ParentSignalHandler` detects and removes it
 during initialization. This prevents a new job from immediately
 checkpointing and exiting due to a leftover trigger from a prior run.
 
+## Manually Triggering a Checkpoint
+
+You can trigger a checkpoint-and-exit without sending a signal by writing
+the trigger file directly. This is useful for debugging, testing, or
+integration with custom orchestration that doesn't use Unix signals.
+
+The trigger file path is:
+
+```
+/dev/shm/instructlab_checkpoint_requested_<JOB_ID>
+```
+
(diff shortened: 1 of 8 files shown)


Maxusmusti (approved):
LGTM

PR #706, opened 2026-06-23
feat: add class-based callback system for training lifecycle hooks

### LINKED ISSUE #694: Add Callbacks to Instructlab Training
## Overview

**Goal:** Add Callbacks to Instructlab-Training 

**Why:**

We want to facilitate use-cases such as:
* initiating eval jobs within a GPU cluster on checkpoint save
* ending the training loop early based on some condition

In order to implement this, we would add features in the following order: 
1. [ ] implement simple, read-only callbacks
2. [ ] implement callbacks that can control flow of training loop (e.g. stopping the loop early, initiating a checkpoint, etc.) 

## Expected usage

Users calling the `run_training` API would directly provide their callbacks to the `TrainingArgs` object through a flat API. 

As an example:

```py
run_training(
  on_checkpoint=run_eval,
  on_validation_loss=end_if_diverged,
  # ... etc
)
```

## Implementation Details

To implement callbacks in this training library, we have to worry about two major pieces:
1. Taking callbacks that were provided to the `run_training` API and passing them to the `torchrun` subprocesses which run the actual training loop
2. Receiving the callbacks inside of the training loop (child process of `torchrun`) and executing them at the appropriate points without interfering with the training process

We will eventually also want to implement callbacks which can modify what actions are performed by the training loop itself, however this will require some additional effort to ensure proper synchronization of the global process group.  These callbacks will not be covered in this issue though. 

### The callback API

We can define a general callback as a generic function which accepts some general context as well as a list of args + kwargs:

```python
def my_callback(context, **kwargs) -> None:
  # do stuff
  return
```

Callbacks may also have arguments for certain events, e.g. `on_save` may have the following interface:

```py
def on_save(context, checkpoint_path: str, **kwargs) -> None:
  # do stuff
  return
```

While the `on_evaluate` callback would look like this:

```py
def on_evaluate(context, validation_loss: float, **kwargs) -> None:
  # do stuff
  return
```

**We expect** these callbacks to be entirely self-contained, so that any and all packages are imported **inside** of the callback function and must be present within whatever venv the training loop runs in. 

Additionally, since we expect the initial callbacks to be **read-only** and inconsequential to the training loop, the training loop will wrap them as asynchronous tasks which run on the global event loop. For this reason, callbacks should be careful with their usage of async functionality. 

> [!warning]
> To prevent any unexpected behavior or complex testing scenarios, the callbacks will not be allowed to have their exceptions propagate into the training loop, therefore any exceptions **must** be handled by the callback or they will fail silently. 

### Adding callbacks to `TrainingArgs` 

To maintain a simple interface, the `TrainingArgs` would expose registration for each callback with a flat interface like this:

```py
from instructlab.training.config import TrainingArgs

args = TrainingArgs(
  on_save=evaluate_gsm8k,
  on_evaluate=save_if_improved,
  ...
)

# then we call training
run_training(args, torch_args)
```

The API for these event hooks should also allow providing multiple callbacks. You would see a definition like this:

```py

class TrainingArgs(pydantic.BaseModel):
  # ...
  on_save: typing.Function | list[typing.Function] | None = Field(None, ...),
)
```

### Passing callbacks to `torchrun` procs

In order to execute callbacks during training, they must first be passed between the main process where `run_training` is called and the child processes that are spawned to run the training loop in parallel. When `run_training` is first called, it uses the `subprocess.Popen` API in order to invoke `torchrun` which then creates a worker for each GPU. The training loop itself expects to be invoked as a CLI, so `run_training` maps all of the fields in the `TrainingArgs` to the CLI flags expected by the actual training script. You can see this in action here: https://github.com/instructlab/training/blob/main/src/instructlab/training/main_ds.py#L624-L801

Therefore we can provide these callbacks to our training script by first serializing them with base64, passing them as string arguments to the CLI, then unserializing and `eval`-ing each callback to register it in the child's memory space. As long as the callback is self-contained, it should execute on the child process just as it would in the main process. 

### Registering and executing callbacks

Once they're unserialized, each callback will be registered **on the global main/rank-0 process only** by a callback manager. 
This is necessary to avoid race conditions or a flood of evaluation requests from each process after every checkpoint save. We can add this functionality later, but we should avoid it to start so the initial implementation can be straightforward. 

In order to manage callbacks, we should have a class like `CallbacksManager` which registers the callbacks and is responsible for providing them with the relevant context when they're called. 

The callback manager would have an `on_<event>` method for each of the following events:
- [ ] on_init_end: Called after backend initialization, before training
- [ ] on_train_begin: Called at the start of training
- [ ] on_epoch_begin: Called at the beginning of an epoch
- [ ] on_step_begin: Called at the beginning of a training step
- [ ] on_substep_end: Called after each gradient accumulation substep
- [ ] on_pre_optimizer_step: Called before optimizer.step() (after gradient clipping)
- [ ] on_optimizer_step: Called after optimizer.step()
- [ ] on_log: Called when metrics are logged
- [ ] on_save: Called when a checkpoint is saved
- [ ] on_evaluate: Called after evaluation
- [ ] on_step_end: Called at the end of a training step
- [ ] on_epoch_end: Called at the end of an epoch
- [ ] on_train_end: Called at the end of training
- [ ] on_predict: Called before prediction/inference
- [ ] on_prediction_step: Called during each prediction step

Internally, the callbacks should be treated as asynchronous coroutines. When the manager's event hook is triggered, it would create a task for each of the registered callbacks and schedule them for execution on the global event loop. There should be no expectation of reading the return values of the callbacks or that they pass values from one to another.

hrathina:
## Summary

Adds a callback mechanism to the InstructLab Training library, enabling users and Training Hub to hook into training lifecycle events without modifying source code.

- Introduces `TrainerCallback` base class with 13 lifecycle hooks that users subclass and override
- Callbacks are async fire-and-forget with exception isolation — they never block or crash training
- Fires on all ranks; each callback gates its own behavior via `context.is_world_process_zero` or `context.is_local_process_zero`
- Serializable across the torchrun subprocess boundary via `inspect.getsource` + base64 encoding
- `on_train_end` blocks up to 10 seconds to allow cleanup callbacks to finish before process exit

### Files changed

| File | Change |
|------|--------|
| `src/instructlab/training/callbacks.py` | **New** — TrainerCallback, TrainingContext, CallbackManager, serialization |
| `src/instructlab/training/config.py` | Add `callbacks` field to TrainingArgs |
| `src/instructlab/training/batch_loss_manager.py` | Add 2 hooks (on_before_forward, on_after_backward) |
| `src/instructlab/training/main_ds.py` | Add 11 hooks in train(), serialization in run_training(), deserialization in main() |
| `src/instructlab/training/__init__.py` | Export TrainerCallback and TrainingContext |
| `tests/unit/test_callbacks.py` | **New** — 34 unit tests |

## Test plan

- [ ] 34 unit tests covering dispatch, snapshot isolation, exception isolation, all-ranks firing, rank gating, kwargs validation, close(), serialization round-trip, public API exposure
- [ ] Full regression suite: 199 passed, 4 pre-existing failures (unrelated LoRA/tensorboard/wandb dependencies)
- [ ] Lint and format clean

Resolves https://github.com/instructlab/training/issues/694

## Summary by CodeRabbit

* **New Features**
  * Added an extensible training callback system with lifecycle hooks around training start/end, epochs/steps, forward/backward timing, optimizer/logging/evaluation, and checkpoint saving.
  * Exposed callback support in training configuration and added CLI support to serialize/deserialize callbacks for distributed runs.

* **Tests**
  * Added comprehensive unit tests covering callback registration/removal, hook dispatch behavior, exception suppression, context snapshot isolation, rank gating, and CLI serialization round-trips.

### DIFF
--- a/src/instructlab/training/__init__.py
+++ b/src/instructlab/training/__init__.py
@@ -5,7 +5,9 @@ __all__ = (
     "LoraOptions",
     "QuantizeDataType",
     "TorchrunArgs",
+    "TrainerCallback",
     "TrainingArgs",
+    "TrainingContext",
     "run_training",
     "FSDPOptions",
     "ShardingStrategies",
@@ -17,6 +19,7 @@ __all__ = (
 import instructlab.training.logger  # Disable package logging by default
 
 # Local
+from .callbacks import TrainerCallback, TrainingContext
 from .config import (
     DataProcessArgs,
     DeepSpeedOffloadStrategy,
--- a/src/instructlab/training/batch_loss_manager.py
+++ b/src/instructlab/training/batch_loss_manager.py
@@ -48,7 +48,14 @@ class BatchLossManager:
     - Computing average losses for logging
     """
 
-    def __init__(self, model, accelerator, world_size: int, local_rank: int):
+    def __init__(
+        self,
+        model,
+        accelerator,
+        world_size: int,
+        local_rank: int,
+        callback_manager=None,
+    ):
         """
         Initialize the BatchLossManager.
 
@@ -57,12 +64,14 @@ class BatchLossManager:
             accelerator: The accelerator instance for distributed training
             world_size: Number of distributed processes
             local_rank: Local rank of the current process
+            callback_manager: Optional CallbackManager for lifecycle hooks
         """
         self.model: Model = model
         self.accelerator: Accelerator = accelerator
         self.world_size: int = world_size
         self.local_rank: int = local_rank
         self.torch_device = torch.device("cuda", local_rank)
+        self.callback_manager = callback_manager
 
     def process_batch(
         self,
@@ -111,6 +120,9 @@ class BatchLossManager:
             batch_total_samples += micro_batch_size
             batch_total_length += total_length
 
+            if self.callback_manager:
+                self.callback_manager.fire("on_before_forward")
+
             # prepare model inputs
             model_inputs = self._prepare_model_inputs(mb)
 
@@ -125,6 +137,9 @@ class BatchLossManager:
                 break
 
             self.accelerator.backward(scaled_loss)
+
+            if self.callback_manager:
+                self.callback_manager.fire("on_after_backward")
 
             # accumulate losses
             grad_accum_steps += 1
--- a/src/instructlab/training/callbacks.py
+++ b/src/instructlab/training/callbacks.py
@@ -171,6 +171,8 @@ class CallbackManager:
             raise ValueError(
                 f"Unknown hook: '{hook_name}'. Valid hooks: {HOOK_NAMES}"
             )
+        if self._loop.is_closed():
+            return
         if not self.has_callbacks(hook_name):
             return
 
@@ -191,8 +193,11 @@ class CallbackManager:
                 TrainerCallback, hook_name
             ):
                 continue
+            cb_snapshot = copy.copy(snapshot)
+            cb_snapshot.batch_metrics = dict(snapshot.batch_metrics)
+            cb_snapshot.val_metrics = dict(snapshot.val_metrics)
             future = asyncio.run_coroutine_threadsafe(
-                self._safe_invoke(method, snapshot), self._loop
+                self._safe_invoke(method, cb_snapshot), self._loop
             )
             if hook_name == "on_train_end":
                 try:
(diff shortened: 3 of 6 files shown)


hrathina:
@RobotSail please review this PR. Thank you.

RobotSail (approved):
Thanks for the PR @hrathina! It seems like the linter is failing, can you double-check this? Otherwise everything looks good to me, LGTM!

efazal:
@Mergifyio queue

Issue #705, opened 2026-05-16
Discussion: testing strategy for instructlab/training as a backend library


gkneighb:
Picking up from the abandoned #404 per @RobotSail's invitation in that thread. The original PR drafted a four-tier testing methodology when this repo was the testing surface for the `ilab` CLI; that framing is stale now that `training_hub` is the user-facing front door and `instructlab/training` is positioned as one of several training backends.

This issue is to open a fresh discussion of what testing in this repo should actually look like given the current scope. Sketching a starting position — please push back, this is not a proposal yet.

### What changed for testing

- **User-facing tests should live in `training_hub`,** not here. Anything that exercises the `run_training()` interface contract from a user perspective belongs at the algorithm-level interface, not the SFT backend.
- **This repo's job is now narrower:** correctness of SFT mechanics on a given backend matrix (FSDP/DeepSpeed × accelerator variant × feature flag set). The integration story is `training_hub`'s problem.
- **Convergence is the harder problem.** Smoke tests can verify "it doesn't crash"; what `training_hub` can't do for us is "this backend actually produces a model whose loss curve looks right on a fixed-seed run."

### Sketched tiers

This is the part most likely to need revision — calling out the shape, not the details.

1. **Linting / type checks.** Pre-test gate. Run on every PR.
2. **Unit tests.** No GPU. Verify isolated mechanics — config serialization, batch metric accumulation, the `BatchLossManager` class contracts, checkpoint round-trip on a CPU-faked model. Always run.
3. **Smoke tests.** GPU-required, per matrix entry. Verify that training runs end-to-end without crashing on a verified configuration list (a handful of named configs that cover the unique code paths — NVIDIA FSDP, NVIDIA FSDP+LoRA, AMD DeepSpeed, etc.). Hard cap somewhere around 30 minutes per entry, target under 10. Block PRs.
4. **Convergence checks.** Per matrix entry, short fixed-seed runs on a fixed dataset, loss curve exported to a stable artifact store, compared against a baseline. The baseline refreshes deliberately on new arch, major dep bump, or training-loop change. Catches the "garbage gradients on accelerator X" case that smoke tests miss.

Downstream model benchmarks (MMLU, MT-Bench) live in `training_hub` or release qualification, not here.

### Open questions

1. **`training_hub` integration tests:** what does that side already cover, and where does it currently break / need this repo's tests to catch failures before integration?
[...]

PR #712, opened 2026-08-24
tox: stop py3-unit re-pinning to Python 3.11, let the CI matrix actually run

UgaTheDev:
`unit.yaml` runs `tox -e py3-unit` across a `matrix.python` of `3.11`, `3.12`,
and `3.13` (three separate jobs, each with `actions/setup-python` installing
the matching interpreter). But `[testenv:py3-unit]` in `tox.ini` sets
`basepython = python3.11`, so `tox` re-pins every leg back to 3.11 regardless
of which interpreter `setup-python` put on `PATH`. All three matrix legs
currently exercise the same interpreter — the 3.12/3.13 legs are decorative.

This removes that pin. With no `basepython` override, `tox`'s `uv-venv-runner`
uses the interpreter already on `PATH` for that job — i.e. whatever
`actions/setup-python` set up for the current matrix leg — so the three CI
jobs actually diverge.

Scope is intentionally narrow: only `py3-unit`. `py3-smoke` (`tox.ini` line 53)
keeps its `basepython = python3.11` pin — it isn't part of the version matrix
in `smoke.yaml`/`smoke-py312.yaml` and shouldn't silently start floating.

Verified locally: `tox -l` still lists `py3-unit` with the same commands;
only the interpreter selection changes. No other env is touched.

Originally flagged in [#705](https://github.com/instructlab/training/issues/705) (comment from 2026-07-29): "the 3.11/3.12/3.13 matrix is decorative: `[testenv:py3-unit]` in `tox.ini` sets `basepython = python3.11`, so tox re-pins the interpreter to 3.11 on all three legs no matter what `setup-python` installed."

Fixes the tox side of the CI-matrix finding in #705.

## Summary by CodeRabbit

* **Chores**
  * Updated the Python 3 unit-test environment to use tox’s default interpreter selection.


UgaTheDev:
All four of these (#712, #713, #714, #715) are mergeable with zero human reviews — and CodeRabbit rate-limited on three of them, so they've had no review at all. Since the constraint looks like review bandwidth rather than the changes themselves: want me to consolidate the four into a single PR so it's one review pass instead of four? They're independent commits, so either shape works — say the word and I'll have it up the same day.

